PR #50, opened 2016-02-05
aliascom fixes

IllusionaryOne:
- moved check for parameters as getting javascript error with args[0] not defined and never making it to usage check: (hook.call, command, ./commands/customCommands.js) TypeError: Cannot call method "replace" of undefined
- aliascom had parameters backwards for commandExists() check and aliases table check.

### DIFF
--- a/javascript-source/commands/customCommands.js
+++ b/javascript-source/commands/customCommands.js
@@ -193,19 +193,20 @@
         return;
       }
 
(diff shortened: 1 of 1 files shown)


ghost:
Is this all tested?

IllusionaryOne:
I had tested this, yes.  Sorry didn't reply originally.

PR #56, opened 2016-02-09
TwitchAlerts Donation First Revision

IllusionaryOne:
Please note that this has been tested without any donations.I do need to test an actual donation. Also, I feel there are extra features to add that I need to still work on but need input:

Add points based on donation - but I would like a calculation from the team for that. Always 1:1? Allow the user to configure X points to 1 dollar? X points to Y dollars configurable?

Presently has !lastdonation -- do we want to shoot for !lastdonation for last 5 donations? Configurable?

Right now I am generating a key for "IllusionaryBot" from www.illusionaryone.com/iobotauth This is fine, but, I am going to assume that the team should host from phantombot.net and register their own application with TwitchAlerts (I can help with this). Also, my scripts do have some requirements, so we should talk about that.

### DIFF
--- a/javascript-source/handlers/donationHandler.js
+++ b/javascript-source/handlers/donationHandler.js
@@ -1,0 +1,110 @@
+/**
+ * donationHandler.js
+ *
+ * Detect and report donations.
+ */
+(function () {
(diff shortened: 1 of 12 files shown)


phantomindex:
what requirements?

PR #105, opened 2016-02-21
Making the whole script much more portable - will work on Mac OS X and FreeBSD now.

shiek:


phantomindex:
failed check

shiek:
failed check? travis build was broken because the sqlite jar was missing o_O

PR #112, opened 2016-02-23
Uglifier Scripts and Various Bug Fixes

IllusionaryOne:
- uglify.bat
  Must be ran from the javascript-source directory, pass a list of files to uglify.
- uglifyall.bat
  Must be ran from the javascript-source directory, will uglify ALL js files.
- uglify.sh
  Must be ran from the javascript-source directory, pass a list of files to uglify or keyword ALL for all files
- core/timeSystem.js
  When getting string for time, 60 seconds was showing as 0sec. Fixed.
- handlers/followHandler.js
  !followreward no longer sets to NaN if no parameters. Usage shows current setting.
- systems/pointSystem.js
  Updated !points setname to support 'delete' option. Cleaned up use of global variables for single and mutilple point names. When reset to original values, was not being assigned back to the global variable.
- lang/english/systems/systems-pointSystem.js
  Created from main.js. Cleaned up usage strings for !points setname
- lang/english/systems/systems-ranksSystem.js
  Cleaned up some of the lang strings
- systems/ranksSystem.js
  Fixed bug with a person with no rank below the lowest rank causing Java String error.
  !rank set was removing the first word in the new rank


IllusionaryOne:
Reverted pointsSystem.js back to original master.  Will try to check in again later.

PR #116, opened 2016-02-24
CustomAPI Support (and minor ranks fixes)

IllusionaryOne:
- scripts/systems/customCommands.js
  - Supports customapi again
  - !addcom (customapi url)
  - This will dump what is read in from a URL to chat.  Users should take caution to review what would come in first.
  - Added new support for customapijson
  - !addcom (customapijson key url)
  - !addcom (customapijson object.key url)
  - !addcom (customapijson object.object...key url)
  - This will perform a simple parse on a JSON object that is returned from a web API. The 'key' is the data to present to the user from the JSON object.  By default, the top level JSON object shall be parsed.
  - The user may choose to dig into a lower level JSON object
  - JSON ARRAYS are NOT supported at this time!  Need to figure out an easy syntax for the user to support instructing the parser as to what is an object and an array.
- scripts/init.js
  - FIX for ranks; stop loading the core/ranks.js (which should be removed)
- scripts/lang/english/systems/systems-ranksSystem.js
  - FIX for ranks; minor updates to lang file.


IllusionaryOne:
Also contains speed-ups for custom commands in general.  The regular expressions used for determining if there are user supplied tags (touser), (sender), etc are now built in advance.  This seems to speed up processing of a single tag like (touser) from 500ms to around 100-200ms.  When testing all tags at once (not that a user would put ALL tags in a command, except for the customapi ones) it takes about 430ms...

[02-24-2016 @ 15:24:35.472] IllusionaryOne: !tags
[02-24-2016 @ 15:24:35.901] [CHAT] |IllusionaryOne|IllusionaryOne|@IllusionaryOne|illusionaryone|IllusionaryBot|bits|Stream is offline|Lifeforce|Testing|2|5|0

PR #173, opened 2016-03-06
Permission (Group) Fixes

IllusionaryOne:
**IrcEventHandler.java**
- Stop MODE +o/-o from sending event to trigger the bot to mod/unmod users as the -o is sent even when modded. Known Twitch issue.

**permissions.js**
- Changed restoreSubscriberStatus() to immediately return if an Admin or Mod is detected.
- Corrected getIniDbBoolean to $.getIniDbBoolean.
- When the groups are loaded, always put Owner in group 0.
- Stop restoreSubscriberStatus() when user leaves the channel.
- ircChannelUserMode event was sometimes treating the boolean true as false and vice versa. Changed to .toString().equals() cast to true.
- Demote those that has been unmodded.
- Perform restoreSubscriberStatus() when a subscriber is detected.


IllusionaryOne:
Do NOT use this copy of permissions.js without the Java build.  Mods can be demoted with the code in place with the old IrcChannelUserModeEvent.java code.  To use this without the build, comment out the primary else statement in $.bind('ircChannelUserMode')

PR #259, opened 2016-03-28
Add local .txt song file functionality to ytPlayer

haltcase:
For your consideration :)

Allows users to enable reading from a locally saved *.txt file of their choosing for now playing information whenever song requests are disabled or otherwise not available. Local now playing has the same functionality as song requests in terms of !currentsong and chat / console announcements.

Commits also include necessary lang additions.

commands:

```
/*** ADDED ***/
!ytp togglelocal - enable or disable local file functionality
!ytp localpath [./path/to/file.txt] - change or view the path to the .txt file. MUST end in .txt ( command checks for this )

/*** MODIFIED ***/
!currentsong - reads from local file if song requests aren't available, responds with "No song is currently playing" if neither are available. As a result this command has moved above the check for the connected player client in order to allow it to run if requests are disabled.
```


citycide:
While using this on stream tonight it turns out it bugs after one update of the local song, so I'm looking into it. Only happened after I lang'd it too :)

BrandenB:
Waiting on someone to test this. And we might merge it in 2.0.8. Not 2.0.7 because we are trying to remove all the issues, and make it as stable as possible.

PR #262, opened 2016-03-29
Uglify on All JS Scripts

IllusionaryOne:
Just making sure all of the scripts are uglified.  Missed a couple of mine and there were a couple of extras.


IllusionaryOne:
Merge Conflict was in Uglify Scripts.

PR #308, opened 2016-04-18
donationHandler - fixed output directory

mihairu:
donationHandler.js use different filepath, than is created in ./addons directory


IllusionaryOne:
Rejecting, looking in GitHub:
./resources/addons/donationchecker
./resources/addons/donationchecker/latestdonation.txt

And it has been there for 3 months.

IllusionaryOne:
Sorry, rejected on accident, will add path to the Github master.

PR #472, opened 2016-05-21
added german language files

maase2:
not finished yet but its a start


BrandenB:
We might merge when its all finished. And make sure the usage and text is translated the same as it would be in English.

ClocxHD:
this would be really nice, if you can integrate it.
But there are a few grammar mistakes. ;)

Zackery:
Didn't mean to close this, sorry! Keep up the good work.

PR #547, opened 2016-06-10
Fix Pricecom - Users not being charged

IllusionaryOne:
**init.js**
- Logic was missing !isMod check so was ignoring all non-mods.


BrandenB:
oops

PR #635, opened 2016-07-08
- Ability to add sound names via the panel

BrandenB:


BrandenB:
Cough @phantomindex :package:

PR #760, opened 2016-08-08
Songrequest Update

lbarnett83:
A more graceful way to handle songrequests + whispermode.


BrandenB:
I don't want to force whisper mode on a command. That's why I added it to work with the normal whisper mode command.

lbarnett83:
My way wasn't forcing, your's was I think :).  Mine will only whisper if whisper mode is on.

BrandenB:
oh I looked at the wrong thing. Sorry.

PR #779, opened 2016-08-12
!addquote [user] [quote...]

lbarnett83:
Quotes can be attributed to a specific user now


BrandenB:
I like this, but I don't want to force users to add a user in each quote. If you can make it so people can still use `!quote [quote]` I will merge this.

PR #784, opened 2016-08-12
Revert "Revert "!addquote [user] [quote...]""

IllusionaryOne:
Reverts PhantomBot/PhantomBot#783


IllusionaryOne:
@lbarnett83 
Re-reverting, but really, this should be a toggle.  Let us say someone wants to add a quote with a Twitch account as the first word and doesn't realize this change is there:
!addquote IllusionaryOne just died...and died and died.

Will add a quote of "just died...and died and died." unless they know the change is there.

Also, with odd named Twitch accounts, you may not mean to, but could end up causing an issue with a quote.  But it can always be fixed by editing it again, both the user and quote message field.

lbarnett83:
If you'd like, I can work that into my fork tomorrow, so you don't have to worry about it.  I'll ensure its all tested before I make a pull request this time :)

PR #786, opened 2016-08-13
!quotemodetoggle

lbarnett83:
1.  Created the !quotemodetoggle command, which allows switching between "!addquote [quote]" and "!addquote [user] [quote]" functionality
2. Re-wrote the addquote function to respect the quoteMode variable.
3. Made the !quotemodetoggle command write changes out to the DB

quoteMode defaults to "!addquote [quote]" if quoteMode doesn't exist in the settings DB


BrandenB:
You're missing a `)` at the end of `getSetIniDbBoolean`

lbarnett83:
And of course by the time this is caught, I'm 2 hours from home. I'll fix it and re-commit as soon as I get home. :)  Cheers Scania, must be nice not to be the only one making typos. ;)

BrandenB:
So this is fully tested with no bugs?

lbarnett83:
I've tested it by switching modes, using !addquote with each mode active, and reviewed the quote getting stored under each mode. I've also shut down and reloaded the bot with each mode active to ensure that mode stays active on a reboot of the bot. If you can think of anything else to test, let me know. Cheers.

BrandenB:
should be fine. Merging.

PR #816, opened 2016-08-23
!host fix

lbarnett83:
fixed the !host command not actually perfroming hosts :+1:


BrandenB:
If you don't mind changing line 154 and line 237 to `args.length < 1`, cause its using `!action` so those are probably broken too.

lbarnett83:
already looking into that :)

BrandenB:
merging thanks!

PR #841, opened 2016-09-05
Monthly points reset, with changable bonus points

lbarnett83:


BrandenB:
I noticed the fix, but I don't want to merge the monthly feature yet, some of the team members are taking a break so it might take a while before I can talk to them. Is there a way you can create a separate pull request for just that fix?

PR #851, opened 2016-09-08
Gambling and YTP precision

ghost:
gambling.js
- As function randRange is working correctly now, range could be set from 1 to 100. Otherwise with (0,100) it has 101 possible outcomes. While it has 100 possible outcomes, you can set winning range from 51-100 and have exactly 50% chance to win. (1-50 and 51-100)

youtubePlayer.js
- For example actual “!nextsong next 2” shows songs in range 1-2 but real programming range starts from 0. So it skips one song. Solution is to decrease minRange by 1 and then create output. It would not be fine to show song’s number 0 so there is “showRange” variable which adds 1 to actual “code range” (the one starting with 0)

### DIFF
--- a/javascript-source/games/gambling.js
+++ b/javascript-source/games/gambling.js
@@ -34,7 +34,7 @@
 			return;
 		}
 
-		if (range < winRange) {
+		if (range <= winRange) {
 			$.inidb.decr('points', sender, amount);
 			$.say($.lang.get('gambling.lost', $.resolveRank(sender), range, $.getPointsString(amount), $.getPointsString($.getUserPoints(sender))));
 		} else {
--- a/javascript-source/games/gambling.js
+++ b/javascript-source/games/gambling.js
@@ -17,7 +17,7 @@
 	function gamble(sender, amount) {
 		var winnings = 0,
 		    winSpot = 0,
-		    range = $.randRange(0, 100);
+		    range = $.randRange(1, 100);
 
 		if ($.getUserPoints(sender) < amount) {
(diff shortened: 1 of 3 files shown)


BrandenB:
Is this fully tested?

ghost:
Yes, I keep using both of these changes.

IllusionaryOne:
Have the latest changes been tested?  Can you provide sample output from the Console/Chat?

Adjusting minRange to 0 seems fine.

ghost:
Here is the sample. http://i.imgur.com/fgR6aeX.png
When minimal range is 0, it rolls exactly 101 numbers (0-100). When minRange is 1, it rolls from 1 to 100 and there can be 50% chance on winning when gamble win range is set to 50. On the sample, I tested it on 60 and all interval values are there screenshoted. That's about gambling system.

To YoutubePlayer file edit-> http://i.imgur.com/CUWV1WK.png
Top part is actual files, the bottom part is with my fix.
In green rectangle I requested 3 songs, but after using command "!nextsong next 5" in the yellow part it shows only song#2 as 1st and song#3 as 2nd song.
Command in blue rectangle "!nextsong" shows correctly song#1 in songrequest playlist.
Next two commands shows song numbers as it is in coding language. So 1st song is #0 and 2nd song is #1. So actually you need to do "!nextsong list 0-5" to show "!nextsong" too. If you do "!nextsong list 1-5" it will skip the 1st song in SongRequest playlist.
After fixing this, comand "nextsong next 5" shows correctly next 5 songs.
- Command "!nextsong list 1-5" shows correctly songs too. Command "!nextsong list 0-5" says "No songs found in that range." but nobody is going to search the list from 0.

PR #871, opened 2016-09-15
!skipsong command update, with settable count for votes needed

lbarnett83:
Also added the !ytp votecount command to set the number of votes needed to skip the current song


BrandenB:
I like this, but can you make it so if the vote count is 0 it defaults to the old way? Also make the vote count 0 by default so it does not break other people's settings. And maybe a toggle to disable the per-user skip message for big channels.

IllusionaryOne:
@lbarnett83 Bump

phantomindex (commented):
@lbarnett83 Bump

IllusionaryOne:
Sorry for the delay, last thing we would request is, if possible, please show some tests that have passed.  It helps us have a warm fuzzy to know that the code and results have been verified.

IllusionaryOne:
Also, please match the current coding style, I believe there are tabs used rather than spaces.  That does not match the current style of the file.  https://github.com/PhantomBot/PhantomBot/blob/master/CODESTYLE.md

_Code specific rules
4 space indents. Make sure your code lines up with the current one or it will be rejected._

We do not say this to be rude, but just to keep everything uniform and looking good <3

IllusionaryOne:
Also, do not take away the ability of the caster to set permissions to their desired level on commands.  I see that you have forced !skipsong to admin only - I mod in a channel with PhantomBot where the mods are often asked to skip songs while the caster is busy.  The default was 1 in the script, but it is up to the caster to change that setting.

Subcommands may be registered with permissions as well, as is the case with '!skipsong now' - please be sure to register that subcommand and not force it to admins only.  It almost sounds like here needs to be a !skipsongvote command that is separate from the core !skipsong command perhaps or !skipsong vote needs to be added and that subcommand could be placed at a lower permission level.

lbarnett83:
fixed all permissions, now operates as:

!skipsong - skips song if user has permission
!skipsong vote - if voteCount is 0, tell user function is disabled, else execute the vote code

my tester asked if it would be better to make it:

!skipsong - if voteCount is 0, tell user function is disabled, else execute the vote code
!skipsong force - skips song if user has permission

would be a fairly simple re-write on my end before I sync.  Let me know please.

_EDIT_ changed the tabs to spaces :)

IllusionaryOne:
Thank you for the adjustments and for helping us keep the style the same!

PR #889, opened 2016-09-21
Remove JCurses

gmt2001:
Remove unused and unnecessary JCurses library and references
Removed abandoned JCurses-based control panel experiment


IllusionaryOne:
@gmt2001 Thanks for removing this!

PR #908, opened 2016-10-01
Guild Wars 2 Module

agaertner:
A very detailed Guild Wars 2 API module with multi-language support.
(german and english included.)

Here an api key to try it out: 
_A9B3F520-5B81-BC4C-B790-0A3952B44A6DCB0C2EAC-11A2-409C-B65D-669B8438A323_
(As caster, set with !gw2 setkey [apikey])

### DIFF
--- a/javascript-source/games/guildwars2.js
+++ b/javascript-source/games/guildwars2.js
@@ -8,6 +8,15 @@
  *
  *          - CHANGELOG -
  *
+ *    V3.0.2
+ *    - Improvements to code.
+ *        Replaced string notations with dot notations for fetching JSON.
+ *        Removed unnecessary semicolons.
+ *        Removed unused variables.
+ *        Fixed scoping for some variables.
+ *        Removed unnecessary URI encoding.
+ *    - Fixed deathcounter for character names with multiple surnames.
+ *
  *    V2.9.7
  *    - !gw2 deathcounter and !gw2 goldcounter can now be called by moderators and above.
  *    - !gw2 build [gamemode] [character] can now also be called with swapped arguments: !gw2 build [character] [gamemode].
@@ -164,32 +173,33 @@
     function CalcCoins(coins) {
         coins = coins || 0;
         var c = coins % 100;
+        var s;
         if (Math.abs(coins) >= 10000) {
             coins = (coins - c) / 100;
-            var s = coins % 100;
+            s = coins % 100;
             var g = (coins - s) / 100;
             c = Math.round(c);
-            if (s.toString().match(/^\s*-?[0-9]\s*$/)) { s = '0' + Math.abs(s) } else { s = Math.abs(s) };
-            if (c.toString().match(/^\s*-?[0-9]\s*$/)) { c = '0' + Math.abs(c) } else { c = Math.abs(c) };
+            if (s.toString().match(/^\s*-?[0-9]\s*$/)) { s = '0' + Math.abs(s); } else { s = Math.abs(s); }
+            if (c.toString().match(/^\s*-?[0-9]\s*$/)) { c = '0' + Math.abs(c); } else { c = Math.abs(c); }
             if (GW2_coinformat) {
                 return g + '.' + s + ',' + c + 'g';
             } else {
                 return g + 'g ' + s + 's ' + c + 'c';
-            };
-        };
+            }
+        }
         if (Math.abs(coins) >= 100) {
             coins = (coins - c) / 100;
-            var s = coins % 100;
+            s = coins % 100;
             c = Math.round(c);
-            if (c.toString().match(/^\s*-?[0-9]\s*$/)) { c = '0' + Math.abs(c) } else { c = Math.abs(c) };
+            if (c.toString().match(/^\s*-?[0-9]\s*$/)) { c = '0' + Math.abs(c); } else { c = Math.abs(c); }
             if (GW2_coinformat) {
                 return s + ',' + c + 's';
             } else {
                 return s + 's ' + c + 'c';
-            };
-        };
+            }
+        }
         return Math.round(c) + 'c';
-    };
+    }
 
     /**
      * @function _getJSON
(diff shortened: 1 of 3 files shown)


IllusionaryOne:
Does not match coding style: https://github.com/PhantomBot/PhantomBot/blob/master/CODESTYLE.md
- 4 space indents. Make sure your code lines up with the current one or it will be rejected.
- Encapsulate all code in (function(){ /\* Your code */ })(); and export globally needed code to $ by function assignment.
- Do not assign variables and functions to the $ global. Do not use the $ api if you variable is local.
- Camel case for variable and function names. (thisIsCamelCase)

Not using helper API functions where they should be used such as:
if (parseInt($.inidb.get("settings", "gw2_formatted")) == 1)

No default permission on the registration of the chat command.  Handling manually within functions.  Casters often want to allow other folks to run chat for them as they are busy streaming.

Ambiguous JavaScript conditions used that do not check the types.

Writing to files with newlines that are not compatible with all OSs.

Calling SaveAll() which is only used by IniStore and forces the bot into a wait state mode if IniStore is used during that operation which is why it is used at shutdown or reload of configuration only.

Not using API commands to open and write files.

There may be more, but this was my first pass.

agaertner:
$.isCaster condition is only used for !gw2 setkey and !gw2 deathcounter !gw2 goldcounter, as those depent on file monitoring on the caster local machine.
I put $.isMod to !gw2 coinsformat. The rest has no restriction and is usable by everyone.
- Encapsulated.
- Removed $ global.
- Camel cases checked (only really needed for conditions, right? I always do it in IF statments.) Checked and added fixed 2 at the top.
- helper API functions removed and now calling them once after initReady, while bot is running, any changes will be saved, but stored in local variable for reuse without needing to load from DB.
  (ie. `var GW2_coinformat = ($.inidb.exists('settings', 'gw2_coinformat') ? parseInt($.inidb.get('settings', 'gw2_coinformat')) : 0);`)

Ambigious Javascript conditions that does not check the types? Do you mean the == undefined ones? Or is , because I am not using ===? Well, it's kinda ok to use == here, when "1" == 1 the string equals the integer. It's ok for my purpose. It will never be something else than an integer parsed inside a string. (I am sure thought, that every string is parsed as integer in my conditions.)

> Writing to files with newlines that are not compatible with all OSs.

That one you gotta explain me how to fix. (Don't know about linux folder structure.)
- Removed SaveAll()

> Not using API commands to open and write files.

That you gotta explain as well. :/

Btw, is there a problem how my @commandpaths are setup? Can the compiler read it like that? Anyway, I changed every @commandpath tag to be right before the function like it is shown in the coding style example.
- Changed variable setup

(not committing before fix for file writing on other operating systems.)

Preview of changes: http://pastebin.com/tJ4VWgzc
(v2.7.9 WIP) WIPChangelog included

Btw 4 space indents? I doN't see a difference in line-up when comparing to files like quoteSystem.js. Maybe I am misunderstanding something or my english is just too bad.

I am always using a tab to encapsulate /\* CODE \* / that belongs into statements or functions such as IF or FOR, DO WHILE. for readability. I really don't see a difference in code line up to ../*system.js

(Added api key in my first post of this pull request to try it out. API keys by ArenaNet only give read-permission.)

IllusionaryOne:
Regarding spacing, this is what your file looks like when I view control characters, tabs are being used, we use spaces, not tabs:
(function(){$
^I// CACHE & DATABASE. DO NOT TOUCH!$
^Ivar GW2_apiKey = ($.inidb.exists('settings', 'gw2_apikey') ? $.inidb.get('settings', 'gw2_apikey') : '');$
^I// Minimum required permissions: account, wallet, characters, pvp, builds, progression.$
^Ivar GW2_apiURL = 'https://api.guildwars2.com';$
^Ivar GW2_coinformat = ($.inidb.exists('settings', 'gw2_coinformat') ? parseInt($.inidb.get('settings', 'gw2_coinformat')) : 0);$
^Ivar GW2_leagues = ['amber', 'emerald', 'sapphire', 'ruby', 'diamond', 'legendary'];$
^Ivar GW2_tiers = ['Ⅰ', 'Ⅱ', 'Ⅲ', 'Ⅳ', 'Ⅴ', 'Ⅵ', 'Ⅶ', 'Ⅷ', 'Ⅸ', 'Ⅹ'];$
^Ivar GW2_professions = ['elementalist', 'engineer', 'guardian', 'mesmer', 'necromancer', 'ranger', 'revenant', 'thief', 'warrior'];$
^Ivar GW2_toggle_goldcounter = 0;$
^Ivar GW2_overall_gold;$
^Ivar GW2_session_gold;$
^Ivar GW2_toggle_deathcounter = 0;$
^Ivar GW2_deathcounterChar;$
^Ivar GW2_overall_deaths;$
^Ivar GW2_session_deaths;$
^Ivar GW2_last_game;$
^Ivar GW2_characters;$
^Ivar GW2_accCreated;$
^Ivar GW2_accAccess;$
^Ivar GW2_accName;$
^Ivar GW2_world;$
^Ivar GW2_world_name;$
^Ivar GW2_world_population;$
^Ivar GW2_commander;$
^Ivar GW2_guilds;$
^Ivar GW2_fractal_level;$
^Ivar GW2_wvw_rank;$
^I// ###############################$
$
    /*_$
     \* @function CalcCoins$
     \* @param {Number} coins$
     \* @returns {String}$
     */$
^Ifunction CalcCoins(coins) {$
^I^Icoins = coins || 0;$
^I^Ivar c = coins % 100;$
^I^Iif (Math.abs(coins) >= 10000) {$
^I^I^Icoins = (coins - c) / 100;$
^I^I^Ivar s = coins % 100;$
^I^I^Ivar g = (coins - s) / 100;$
^I^I^Ic = Math.round(c);$
^I^I^Iif (s.toString().match(/^\s_-?[0-9]\s_$/)) { s = '0' + Math.abs(s) } else { s = Math.abs(s) };$
^I^I^Iif (c.toString().match(/^\s_-?[0-9]\s*$/)) { c = '0' + Math.abs(c) } else { c = Math.abs(c) };$
^I^I^Iif (GW2_coinformat) {$
^I^I^I^Ireturn g + '.' + s + ',' + c + 'g';$
^I^I^I} else {$
^I^I^I^Ireturn g + 'g ' + s + 's ' + c + 'c';$
^I^I^I};$

agaertner:
Oh this is how you meant it. Now I understand. But why are tabs bad? Or is it just because the team decided on spaces? Tabs are so much faster D:

IllusionaryOne:
Because there is a coding style, and yes, the team decided upon 4-space indents.  Most projects would have a coding style defined.  Everywhere that I have worked has had coding style documentation as well, sometimes near 100 pages in length depending upon how many languages were used.

agaertner:
Ah, at first I thought it because tabs are bad technical wise.. ok.   So I am gonna replace 1 tab with 4 spaces. edit: 1971 occurences of a tab were replaced with 4 spaces. I think I got them.

Nekress:
Do you know of a way to make writeToFile compatible with the other platforms?

agaertner:
Today I started my subject in computer science at my university and we were told that they mostly work with Linux and are using GIT for versioncontrol. (they also told us about Subversion etc.) Anyway, I asked what the difference in folder structue on Linux is compared to Windows as I never used it before (told them I was working on something that needed compatibility for several OS.) I learned that Linux does not have storage drives, but only one and C:\User\ on windows is /home/{username} on Linux and external drives are /media/{username}/USB.... Well that didn't help me with this problem (since those technical paths are defined by the methods), but I will look into the followHandler.js and fix it.   I am quite motivated to learn more stuff about this! This is exciting! I also didn't know files in Linux are case-sensitive! foO.txt is a different file than Foo.txt. (On windows its the same.)

EDIT: Apparently, I only forgot to add ./ to the beginning of the path in writeToFile();
./addons/

itsDave1:
Just a tip, there's a lot of errors in your code. Here's a screenshot of a few.

image

agaertner:
Cool page (JSHint.com), I parsed it several times and fixed everything!
Only returns false-positives now for "$", "Packages" and "java".

I bookmarked this web tool. thanks.
In return, I give you this [JSON validator](http://jsonlint.com/) with a cool [pro page](http://pro.jsonlint.com/). (settings on the right.)
and [JSON parser](http://json.parser.online.fr/beta/) (with settings gear top right corner.)

agaertner:
I think JSHint has one problem though on the rror "var x is already defined" when in reality, the variable is separated by two IF-statements that would return and can never be called both in one run.
For example:

```
function example(number) {
    if (number >= 1000) {
        var x = someValue;
    return;
    }
    if (number >= 100) {
       var x = anotherValue;
    return;
    }
}
```

When number is equal to or over 1000 the second condition doesn't get called, because the first one gets returned, stopping the function before it reaches the second condition. Therefore variable x IS NOT "already defined" (not trying to double define) as JSHint.com tries to tell me.

Anyway, I put `var x;` above both conditions and removed the `var` from the lines so JSHint doesn't give me that error, even though I really believe it is a false-positive.

itsDave1:
`var x is already defined` is not an error. Javascript is a non-scoped language. Making variables available wherever, as long as there is nothing escaping the code. This is why you always need unique variables.

BrandenB:
I'm going to close this for now.

PR #927, opened 2016-10-11
databaseCommand.js - Debugging plugin for end-user.

agaertner:
A quick debugging plugin for phantombot.

Why?

When a developer (core or plugin dev) makes changes to his/her database tables or keys he/she can tell the users to call !dbedit [action] [args] to clear **old obsolete** keys or tables so they don't conflict with the db changes in a new release of his/her script.
- It will warn the user if table/key exists and prompt him to reply with 'yes' or 'no' before a change is made permanent.
- The user has 20 seconds to reply and cannot call another subcommand for the duration or before a valid respond is given.

### DIFF
--- a/javascript-source/commands/databaseCommand.js
+++ b/javascript-source/commands/databaseCommand.js
@@ -1,0 +1,117 @@
+/*
+ * Copyright (C) 2016 phantombot.tv
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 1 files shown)


BrandenB:
this is already a debug command in the core.

IllusionaryOne:
While this is a neat addition, this is something never meant to be ran from Chat.  It is meant to only be executed from the Console.  It could be dangerous to let Moderators run such commands, this is why you have Production DBAs in the real world that understand the data sets and relationships.

agaertner:
I could make an exception array (a blacklist) including all the core tables that can't be used in !dbedit.

So it would just be a means for plugin makers.

phantomindex:
Honestly, I don't see the point. If you want this for your bot then that is fine. We don't think it should be a main part of the bot. 

My reason is if someone who's new wanted to see what those commands do without looking, it may change something they never intended to change. 

I also must say the same for GW2 module, it's too specific, our bot is meant for the general public, not to specialize on one game.

I would like to one day have a page where plugins such as yours could be added to a list where anyone could download and throw in their bot, similar to Bukkit for Minecraft.

PR #930, opened 2016-10-15
Update build.xml

stachu540:
Property changes. Now properties got location value. There is no need to setting value. (Tested by myself in _Ant 1.9.6_) Any other Ant builder (lower version) attached from IDE will worked for this changes.


IllusionaryOne:
Brought down this branch and tested with a compile as well.

PR #943, opened 2016-10-26
Core Notice Timer Updates

IllusionaryOne:
**NoticeTimer.java**
- Updated to include a default timer which will just cycle through the notices.
- Handle the stream being offline and handling the notice toggle for that situation.


IllusionaryOne:
This should finalize the primary Core changes, other than enabling the Notice system.  Need to complete the module and Control Panel changes.

PR #954, opened 2016-11-01
Notification when adventure is ready & on cooldown

polarkac:
Small changes so the bot notify when the new adventure is ready and also when the adventure is still on cooldown.

### DIFF
--- a/javascript-source/games/adventureSystem.js
+++ b/javascript-source/games/adventureSystem.js
@@ -28,7 +28,9 @@
         currentAdventure = 1,
         stories = [],
         moduleLoaded = false,
-        lastStory;
+        lastStory,
+        isReady = true,
+        lastStoryTime = 0;
 
 
     function reloadAdventure () {
@@ -223,6 +225,14 @@
      * @returns {boolean}
      */
     function joinHeist(username, bet) {
+        if (!isReady) {
+            var date = new Date();
+            var time = date.getTime();
+            var delta = Math.round(coolDown - (time - lastStoryTime) / 1000);
+            $.say($.whisperPrefix(username) + $.lang.get('adventuresystem.adventure.cooldown', delta));
+            return;
+        }
+
         if (currentAdventure.gameState > 1) {
             if (!warningMessage) return;
             $.say($.whisperPrefix(username) + $.lang.get('adventuresystem.join.notpossible'));
@@ -281,6 +291,7 @@
             line,
             t;
 
+        isReady = false;
         currentAdventure.gameState = 2;
         calculateResult();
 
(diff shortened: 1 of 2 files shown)


BrandenB:
I won't merge this since it does no use the normal cooldown script, and it forces a cooldown message and the ready message has no toggle.

polarkac:
It doesnt use normal cooldown script because it would have to be outside adventureSystem.js or required some changes to cooldown script.
I can add toggle on/off switch, but its pointless due to the cooldown script requirement. If you can point me to the right direction, how to handle the cooldown properly I can give it a shot and implement it.

IllusionaryOne:
Without staring at this too much, couldn't you use the $.cooldown.get() method, in combination with the 'adventure' command to get the remaining cooldown value?  If you are looking for how much longer a command is to be cooled down, I believe that method performs that operation.  

You can then set an interval timer on the $.cooldown.get() method to see when it returns 0 to determine when adventure is available again, right?  Perhaps have it check every 10 seconds or so.  Then set a local variable indicating that you told chat that adventure was available so that it doesn't print out the message again.  Then, when someone runs adventure, you reset that variable and it repeats when the aforementioned method returns 0.

Then, you are not modifying outside of adventureSystem.js and you are using the API that is available for the cooldown as it is meant to be used.

Then yes, the toggle would be a good idea, not all folks want a line presented in chat indicating that the adventure is open.  Also, in my mind, this needs to be configurable via the Control Panel as an option, not just from chat.

polarkac:
Well, the notification when adventure is ready is one thing. Thats quite simple with just an timeout triggered after the same cooldown time as the command. In this case I can use $.cooldown without problem.

Main problem is notification for !adventure command itself. When viewer send the command, bot should inform that adventure is on cooldown and how much is remaining. $.cooldown completely block the command, so it does not trigger any logic. I would have to make some logic outside the adventureSystem.js to be able to get remaining time and send the message.
I dont think this is a good idea or at least its not clean way to do this in my opinion.

BrandenB:
People can toggle if they want the message for the cooldown. `!botname togglecooldownmessage`.
For big channels with over 1000 viewers, if 100 people type the command at the same time not everyone is going to want the cooldown message to show up 100 times.

IllusionaryOne:
Not only that, there are limits to how many times a bot is allowed to chat on Twitch, there are rate limits, and the bot is self-limited.  I am in plenty of decent sized channels with 100-500 people where the "adventure" like command says nothing - doesn't even tell people they entered - more than likely for this reason as well as what Scania has brought up.  That is why it must be an option.  Most users on Twitch have an idea of what a cooldown is on a "adventure" type command.

When submitting code to PhantomBot, think not only of _your_ channel but the channels of _all_ PhantomBot users.  If that does not interest you, then, your option is to create a custom/ directory and keep a separate instance of scripts that are not overwritten during a release.  I do not say this to be mean spirited, obtuse, nor overly critical; this is something that I have had to come to terms with as well as a bot developer.

BrandenB:
I'm going to close this for now.

PR #1045, opened 2016-12-07
Change target and source compile.

stachu540:
Change target and source compile, when Java 7 support has been deprecated.


BrandenB:
Thank you. :)

PR #1061, opened 2016-12-09
Remade the button(s) to enable and disable modules for superior UX

strongdoctor:
I've had issues for a long time now with the enable/disable buttons. Any new users I've introduced to PhantomBot just haven't understood what the little circles do(and that they're clickable at all).

### DIFF
--- a/resources/web/panel/css/style.css
+++ b/resources/web/panel/css/style.css
@@ -805,3 +805,66 @@ input, textarea {
     border-color: grey;
 }
 
+.togglebutton {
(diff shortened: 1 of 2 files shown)


strongdoctor:
Also, this is how it looks currently: http://i.imgur.com/raV9CmI.png ; and the "ball" turns into a spinner that pulses(fa-pulse) when you toggle a module, but due to browsers rendering text different it "wobbles" when spinning around a little bit.

PR #1062, opened 2016-12-10
Revert "Remade the button(s) to enable and disable modules for superior UX"

BrandenB:
Reverting because this because it was merged on accident. 

Thanks, PhantomIndex.


IllusionaryOne:
I will be downloading and testing the code and then we will discuss internally.  One item is, after discussing with other colleagues outside of PhantomBot, is the concept of a UI having one look and feel.  This looks great, but that takes away from the look and feel.  We had an Enable/Disable button in the past but opted for consistency.  Will test and post results.

strongdoctor:
Lmao no worries @ScaniaTV 

@IllusionaryOne 
I've already started work on replacing *all* toggle buttons with the toggle button I implemented in my fork of PhantomBot. What I did just now was finish up the other kind of toggle button, for example used under "Commands -> Permissions":
Image

Personally I don't see how these buttons negatively affect the look and feel, but then again I don't know your plans for the project.

PR #1079, opened 2016-12-17
issue error correctly class

stachu540:
Imported class have same name as object in current class.

If kinda help you can merge it. It's up to you.


BrandenB:
Thank you.

PR #1099, opened 2016-12-25
Added option to allow bits rewards to be a multiplicative factor of the number of bits

EmptyM:
- Added !bitsrewardmulttoggle which allows the reward to become a multiplicative factor of the number of bits instead of a constant amount
- Added toggle to web panel
- I think "!isNaN(action)" should have been "isNaN(action)" in the if statements for !bitsreward and !bitsminimum
- Fixed a bug where !bitsreward 0 and !bitsminimum 0 wouldn't work as !parseInt(action) with a value of 0 would cause the usage message to print. Using isNaN(parseInt(action)) instead.

### DIFF
--- a/javascript-source/handlers/bitsHandler.js
+++ b/javascript-source/handlers/bitsHandler.js
@@ -5,6 +5,7 @@
  */
 (function() {
 	var toggle = $.getSetIniDbBoolean('bitsSettings', 'toggle', false),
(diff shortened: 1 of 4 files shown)


BrandenB:
One small thing that I am seeing is that you're not parsing `event.getBits()` which is a string that the core returns, so it may cause issues later on. Line 51 and 62

EmptyM:
I see. Updated

PR #1100, opened 2016-12-25
Added TwitchAPIv5 and updated other classes to use it

gmt2001:
Methods accepting String-based channel/user names left in for legacy support
String-based channel/user names are automatically converted to v5 user ids
String to ID conversions are cached in memory until TwitchAPIv5 object is destroyed (shutdown)


gmt2001:
P.S. Please give it a good test before merging in case the Twitch Developer Docs made me miss something

Additional commit note: If a Client ID has not been set in the instance of TwitchAPIv5, attempting to make a request will throw an IllegalStateException (will be caught and returned as a JSONObject)

BrandenB:
Going to try to test this out tomorrow if I get a bit of free time. Will also ask the other guys to see what they think.

gmt2001:
Honestly, if you guys want to start a brand new class from scratch to handle this that's probably a good idea, but at least if this one is working properly it will allow you to not worry about the 2018 deadline so you can take all the time that's needed to get it right.

BrandenB:
Sorry for the late reply, but I think that's what we are going to do. Thank you for your effort in doing this, and I will leave this pull-request opened for now in case we do need it.

BrandenB:
Going to close this now, we'll be working on the new class soon.

PR #1101, opened 2016-12-26
Added XML-based lang system for HTML and JS files in the web folder

gmt2001:
Converted 2 phrases to lang system as a test/demonstration

### DIFF
--- a/resources/web/index.html
+++ b/resources/web/index.html
@@ -33,7 +33,7 @@
         </div>
 
         <div class="main">
-            <div class="launcher" data-toggle="tooltip" data-placement="right" title="The username and password are found in botlogin.txt, see the paneluser and panelpassword settings.">
+            <div class="launcher" data-toggle="tooltip" data-placement="right" title="<lang index_tooltip_userpass_location />">
                 <script>
                     var urlInfo = window.location.host.split(":");
                     var url = urlInfo[0];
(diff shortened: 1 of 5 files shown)


gmt2001:
This allows adding lang support to html and js files in the web folder. The translations are stored as XML files in `web/lang/<langname>`.

This is only supported by `.html` or `.js` files that are served by NEWHTTPServer or NEWHTTPSServer.

To add a lang replacement to an html or js file, just put the placeholder `<lang replacement_id />` where `replacement_id` matches the `id` attribute of an entry in the XML files. Note that the space after `<lang` and before `/>` are important. Lang id's can contain letters, numbers, hyphens `-`, underscores `_`, and periods `.`. Matching capitalization is required.

Note that unlike the lang system for the Rhino scripts, this does not support replacement tokens for variables, just literal text.

Like the Rhino lang system, English is always loaded as the default in case the selected language lacks an entry, but any entry, even blank will still count as valid.

The first time the http server runs the lang system, it will load the current lang from the DataStore if it has been set. This lang system subscribes to IrcChannelMessageEvent and will listen for `!lang` commands issued by a moderator (as determined by IRCv3 Tags). Because of this, it is important that lang folder names for the web side match the folder names for the Rhino side.

As an example (and a test, please test it) 2 existing strings in `web/index.html` have been replaced with lang translations. They are the `The username and password are found in botlogin.txt, see the paneluser and panelpassword settings.` tooltip and the `Launch PhantomBot YouTube Player` link text.

PR #1108, opened 2016-12-27
- Added option to allow resub rewards to be separate from sub rewards

EmptyM:
- Created a toggle on control panel as well as a command to allow the reward to be separate
- Created input on control panel as well as a command to update resub rewards
- Also Fixed a bug where resubwelcometoggle was updating subWelcomeToggle instead of reSubWelcomeToggle

### DIFF
--- a/javascript-source/handlers/subscribeHandler.js
+++ b/javascript-source/handlers/subscribeHandler.js
@@ -10,7 +10,6 @@
         subWelcomeToggle = $.getSetIniDbBoolean('subscribeHandler', 'subscriberWelcomeToggle', true),
         primeSubWelcomeToggle = $.getSetIniDbBoolean('subscribeHandler', 'primeSubscriberWelcomeToggle', true),
         reSubWelcomeToggle = $.getSetIniDbBoolean('subscribeHandler', 'reSubscriberWelcomeToggle', true),
-        reSubRewardToggle = $.getSetIniDbBoolean('subscribeHandler', 'reSubscriberRewardToggle', false),
         subReward = $.getSetIniDbNumber('subscribeHandler', 'subscribeReward', 0),
         reSubReward = $.getSetIniDbNumber('subscribeHandler', 'reSubscribeReward', 0),
         customEmote = $.getSetIniDbString('subscribeHandler', 'resubEmote', ''),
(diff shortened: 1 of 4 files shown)


javascript-source/handlers/subscribeHandler.js
```diff
@@ -100,21 +102,30 @@
...
             }
-
             $.say(message);
             $.addSubUsersList(resubscriber);
             $.restoreSubscriberStatus(resubscriber, true);
             $.inidb.set('streamInfo', 'lastReSub', resubscriber);
-            if (subReward > 0) {
-                $.inidb.incr('points', subscriber, subReward);
-            }
+            if (reSubRewardToggle) {
```
BrandenB: This toggle is not needed, you can just do:

```
if (reSubReward > 0) {
       $.inidb.incr('points', resubscriber, reSubReward);
  }
```

javascript-source/handlers/subscribeHandler.js
```diff
@@ -154,6 +165,15 @@
             $.setIniDbBoolean('subscribeHandler', 'reSubscriberWelcomeToggle', reSubWelcomeToggle);
             $.say($.whisperPrefix(sender) + (reSubWelcomeToggle ? $.lang.get('subscribehandler.resub.toggle.on') : $.lang.get('subscribehandler.resub.toggle.off')))
         }
+
```
BrandenB: This toggle is not needed. Just set the reward to 0 by default.

javascript-source/handlers/subscribeHandler.js
```diff
@@ -202,13 +222,29 @@
...
-            $.say($.whisperPrefix(sender) + $.lang.get('subscribehandler.reward.set'));
+            $.say($.whisperPrefix(sender) + $.lang.get('subscribehandler.sub.reward.set'));
+
+        }
+		
+		/**
+         * @commandpath resubscribereward [points] - Set an award for resubscribers.
+         */
+        if (command.equalsIgnoreCase('resubscribereward')) {
+            if (args.length === 0) {
```
BrandenB: Use `action === undefined` here please.

javascript-source/handlers/subscribeHandler.js
```diff
@@ -202,13 +222,29 @@
...
+                $.say($.whisperPrefix(sender) + $.lang.get('subscribehandler.resub.reward.usage'));
+                return;
+            }
+            $.inidb.set('subscribeHandler', 'reSubscribeReward', parseInt(args[0]));
+            reSubReward = parseInt(args[0]);
+            $.say($.whisperPrefix(sender) + $.lang.get('subscribehandler.resub.reward.set'));
+            $.log.event(sender + ' changed the reSubscriber reward to ' + reSubReward);
```
BrandenB: The log event and the return is also not needed.

javascript-source/handlers/subscribeHandler.js
```diff
@@ -100,21 +102,30 @@
             }
 
             if (message.match(/\(reward\)/g)) {
-                message = $.replace(message, '(reward)', String(subReward));
+                if (reSubRewardToggle) {
```
BrandenB: Don't need a toggle here. it can just be `message = $.replace(message, '(reward)', String(reSubReward));`

EmptyM:
So just use reSubReward if its greater than 0, and if not use subReward?

BrandenB (commented):
Most of this is good, just fix the few things I've noticed and I'll merge it after.

BrandenB:
If there is not reSub reward set then the caster probably doesn't want one. So I would just make it not give a reward.

BrandenB:
Perfect! Thank you.

PR #1123, opened 2017-01-02
Added option to allow adventure to only work when channel is online

EmptyM:
- Fixed bug where adventure cooldown message didn't show
- Updated cooldown message to show units (seconds)


javascript-source/core/commandCoolDown.js
```diff
@@ -103,6 +103,7 @@
             if (command.equals('adventure')) {
                 if (cooldown[command] !== undefined) {
                     if (cooldown[command].time - $.systemTime() >= 0) {
+                        $.say($.whisperPrefix(username) + $.lang.get('adventuresystem.adventure.cooldown', parseInt(parseInt(cooldown[command].time - $.systemTime())/1000)));
```
BrandenB, line 106: We do not want to force a cooldown message, plus `!yourbotname togglecooldownmessage` will enable the cooldown message for all command. We do not really want to have a bunch or cooldown messages for X command with a toggle either.

PR #1127, opened 2017-01-02
Fix for raffle concurrency issues.

Heerkog:
Fix for issues discussed at
https://community.phantombot.tv/topic/1228/rafflesystem-entrycount/16


BrandenB:
#1134 should fix this.

PR #1148, opened 2017-01-12
Added currency code support for Streamlabs API

gmt2001:
Defaults to current behavior (Use currency of the donation)


IllusionaryOne:
Once in a while, due to timing, the DataStore is not fully initialized at that point.  Could the StreamLabs API call be moved into init() around where the Twitter initialization is done?  I know this sounds silly, but, in the past, occasionally, I have seen issues trying to read the DataStore right after the object is created which is right before the current StreamLabs call.  If not, I will merge and make the change, not a  big deal.

Cheers

IllusionaryOne:
Thanks!  Merged!

PR #1149, opened 2017-01-14
Allow specific commands to have explicitly chosen cooldown types

DecibillyJoel:
!cooltype [command] [type]
This command allows you to override PhantomBot's logic for figuring out which type of cooldown to use for a specific command. This is especially useful for scenarios where you have a global cooldown time but you want to change a few commands to have no cooldown, or if you want to enable both global and per-user cooldowns but you want to specify some commands' global cooldown times.

The four types of cooldown "types" are:
A) default - this is the default setting, and allows PhantomBot's logic to figure out what type of cooldown each command follows
B) per-user - this overrides PhantomBot's logic and enforces per-user cooldowns for whatever commands you give this type to
C) global - this overrides PhantomBot's logic and enforces global cooldowns for whatever commands you give this type to
D) none - this overrides PhantomBot's logic and enforces no cooldowns for whatever commands you give this type to

Example Scenarios:
A)
    !toggleglobalcooldown
    !toggleperusercooldown
    !globalcooldowntime 2
    !cooltype !8ball global
    !coolcom !8ball 30

B)
    !toggleglobalcooldown
    !globalcooldowntime 2
    !coolcom !slot none

C)
    !toggleperusercooldown
    !cooltype !8ball global
    !cooltype !slot global
    !globalcooldowntime 10

The main reason I did this was because I noticed that with the way PhantomBot's logic currently works, it is impossible to specify specific commands to have a specific global cooldown time (AKA not the generic !globalcooldowntime) if you have both per-user and global enabled. There are some other scenarios that this !cooltype concept fixes as well, but I won't get into listing all of them since the above case should be enough to merit this change.

### DIFF
--- a/javascript-source/core/commandCoolDown.js
+++ b/javascript-source/core/commandCoolDown.js
@@ -8,17 +8,57 @@
  */
 
 (function() {
+    /**
+     * @function getAllCooldownTimes
+     * @return object
+     */
+    function getAllCooldownTimes(){
+        var cooldownTimes = [];
+        var tableName = 'cooldown';
+        var keyList = $.inidb.GetKeyList(tableName,'');
+        var key;
+        for (i in keyList) {
+            key = keyList[i];
+            cooldownTimes[key] = parseInt($.inidb.get(tableName,key));
+        }
+        return cooldownTimes;
+    }
+    
+    /**
+     * 
+     * @function getAllCooldownTypes
+     * @return object
+     */
+    function getAllCooldownTypes(){
+        var cooldownTypes = [];
+        var tableName = 'cooldownType';
+        var keyList = $.inidb.GetKeyList(tableName,'');
+        var key;
+        for (i in keyList) {
+            key = keyList[i];
+            cooldownTypes[key] = parseInt($.inidb.get(tableName,key));
+        }
+        return cooldownTypes;
+    }
+    
     var globalCooldown = $.getSetIniDbBoolean('cooldown', 'globalCooldown', false),
         perUserCooldown = $.getSetIniDbBoolean('cooldown', 'perUserCooldown', false),
         globalCooldownTime = $.getSetIniDbNumber('cooldown', 'globalCooldownTime', 90),
         modCooldown = $.getSetIniDbBoolean('cooldown', 'modCooldown', false),
         cooldown = [],
         cooldowns = [],
-        cooldowntypes = ["default","per-user","global","none"],
-        i;
-        for(var l = 0;l < cooldowntypes.length;l++){
-            cooldowntypes[cooldowntypes[l]] = l;
+        cooldownTypeEnum = ['default','per-user','global','none'],
+        i,
+        cooldownTypes = getAllCooldownTypes(),
+        cooldownTimes = getAllCooldownTimes();
+        
+    for (i in cooldownTypeEnum) {
+        i = parseInt(i);
+        if(cooldownTypeEnum.hasOwnProperty(i)){
+            cooldownTypeEnum[cooldownTypeEnum[i]] = i;
         }
+    }
+    
     /**
      * @function reloadCooldown 
      */
@@ -32,6 +72,7 @@
     /**
      * @function permCheck 
      * @param username
+     * @param isMod
      * @return boolean
      */
     function permCheck(username, isMod) {
@@ -40,11 +81,11 @@
 
      /**
      * @function commandCheck 
-     * @param command
+     * @param cmd
      * @return boolean
      */
     function commandCheck(cmd) {
-        return (cmd.equals('bet') || cmd.equals('ticket') || cmd.equals('tickets') || cmd.equals('bid') || cmd.equals($.inidb.get('raffle', 'command')) || cmd.equals('adventure'));
+        return (cmd.equals('bet') || cmd.equals('ticket') || cmd.equals('tickets') || cmd.equals('bid') || cmd.equals('adventure') || cmd.equals($.inidb.get('raffle', 'command')));
     };
 
     /**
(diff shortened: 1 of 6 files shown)


phantomindex:
This is pretty good, just give our team a few hours to give their thoughts.

DecibillyJoel:
Wow, such a quick response! And alright, will do.

BrandenB:
I've denied this feature a few times for the simple reason of it will slow down all commands since we need to add extra checks and more database calls. The only way I would accept this is, if we would store all commands with their cooldown types and time in an object in memory so we do not need to call the database, which is basically a rewrite of this entire system.
Also, a lot of your code does not follow our general code style, and the cooldown type command is not in the language files. The option is also not available in the control panel.

DecibillyJoel:
I can easily change it so that the cooldown types are stored in memory, but I need a bit of help with doing that. Does your API have a way to either iterate over the entire table or just run an SQL query? If so, what is the method to do so? What I'm thinking to do is iterate over the table on startup and store it all in memory, then just write to the db whenever the change is needed.

I'm not really sure what is meant by your general code style. Can you show me an example of how I can fix my code so that it matches your style?

I don't know how to add the command to the language files, but if you're willing to tell me how, I'll gladly do it.

I haven't even touched the web panel side of the code, so I could give it a look. I'll probably be able to figure that bit out myself, but if not I'll be sure to ask for help.

BrandenB:
You can just run an SQL query, once you do a query that data should be cached, but I've seen it not happen from time to time making everything slow, especially in bigger channels.

By general code style I mean trying to make things look like they already do in other modules and follow [this guide](https://github.com/PhantomBot/PhantomBot/blob/master/CODESTYLE.md) as well.

Since the panel cannot call the scripts or modify anything in the scripts having a way to update an object from it is a real pain, that's the main reason I haven't added this myself and why we don't store command costs and permission in an object as well.

As of right now, the cooldown system takes about 0-1 millisecond to complete, if it can be the same with this option then I would be happy to merge it.

I'll see if I can't think of a way to make this work without slowing down anything.

DecibillyJoel:
Where can I find documentation for the $ API? If there is none, what is the function for running an SQL query?

phantomindex:
docs.phantombot.tv

DecibillyJoel:
Ah, thanks!

DecibillyJoel:
I've gotten everything to work, but I'm having trouble with the web panel. I can't get my changes to take place, even though when I go to the file through /panel/cooldownPanel.html in the web browser, it returns the same html file as the one in my folder. How do I go about changing the web panel? If needed, I can update my fork to what I currently have.

BrandenB:
That file is not used, everything for the cooldown stuff is located in the commands.html/js file.

BrandenB:
I've also just noticed an issue in your new logic, the global cooldown is not meant to be specified on a command, it it meant to be so you don't have to set a cooldown on every command, instead it will default to the global cooldown time. If you set the cooldown on a command to 0 seconds it will not go into cooldown at all. Once you toggle on the per-user cooldown, the bot will make commands that don't have a cooldown on them per-user and not for every user, the global cooldown will still work as it is intended to.

This logic cannot be changed at this point, or it will screw over a lot of people.

DecibillyJoel:
I'm not sure I understand. My code change does not have any effect on anything until you actually use the !cooltype command. When you do, it allows you to set a global cooldown for specific commands. But if you leave the cooldown types to default (which is the default option), the code acts exactly the same as it used to. Or, at least, this was my intent.

BrandenB:
Guess I misread your first comment, but this could get confusing to some users, since by default it has the same logic as we currently do, but once you start setting a type it changes completely. We like to try to keep things as simple as possible for the most part.
Just being able to specify either a per-user cooldown or a cooldown for everyone on a command would be the simplest way and fastest probably.
I'm not saying you have to do this, it is just what I think would be best for the bot. This system might be getting an update in the future, at the moment we are just trying to get everything super stable and update the panel logic a bit.

DecibillyJoel:
"Just being able to specify either a per-user cooldown or a cooldown for everyone on a command would be the simplest way and fastest probably."

I must've not explained myself properly, I apologize. That quote is what using my !cooltype command does, and what your current logic does not allow when both global and per-user cooldowns have been enabled.

If you don't use !cooltype on a command, or use !cooltype command default, that command gets treated with your bot's current logic.

If you use !cooltype command none, that command will never have a cooldown until its type is changed.

If you use !cooltype command per-user, that command will always be treated with per-user cooldowns. For determining the cooldown, it uses the current per-user logic.

If you use !cooltype command global, that command will always be treated with global cooldowns. For determiming the cooldown, it will use the current global logic.

I'm sorry for the confusion, I should've explained this better in my initial comment. All !cooltype does is override the logic for determining what type of cooldown to use for a specific command.

I've gotten most of the code to work, I just need to fix some small aesthetic issues in the web panel. Unfortunately I'm out right now so I'll have to fix it when I get back. My new(est) code has addressed all of the issues you brought up about the initial pull request.

PS: Thanks for creating that DEVSETUP file, it helped a lot to be able to edit everything in Netbeans instead of Notepad++

DecibillyJoel:
I've finished my code changes. The newest commit includes:
-Fixing scripts/core/commandCoolDown.js from sucky constant DB queries to intelligent memory holding of data and only making DB queries when changes need to be made.
-(Still) adding !cooltype command to allow for more flexibility with cooldowns
-Foreseeing and fixing any problems relating to the commands interface on the webpanel that would arise from fixing scripts/core/commandCoolDown.js
-Adding !cooltype to the lang files
-Adding !cooltype to the webpanel
-Fixing unmatching code styles

Let me know if I need to fix anything else!

DecibillyJoel:
OH! I just realized what that flaw you were talking about was. My bad, sorry. One second, I'll fix it.

DecibillyJoel:
Alright, I've fixed the flaw. Using !cooltype !somecommand global will no longer erroneously force the global cooldown onto the command. Instead it will now properly check first that it does not have its own cooldown before using the global cooldown.

DecibillyJoel:
Maybe it'd also be helpful to add an optional argument to !coolcom / !cooldown to specify the cooldown type. That would help simplify the process a bit. What do you think?

DecibillyJoel:
I don't mean to sound pushy, but what's the status with this? I haven't heard from you guys in a while. Do I still need to change anything?

BrandenB:
I don't think we will be including this in the main build. One of the reason is it's a bit too complex just for a cooldown system, also our current cooldown system isn't that fast, so adding more to it will slow down the performance of all command, which we do not want to do. Feel free to submit this to the custom modules repo and if people want it they can add it to their bots.

DecibillyJoel:
Well, you asked me to change the code so that your cooldown system would be sped up, which is what I did in addition to just the cooldown type command. Would you be willing to accept the PR if I just removed the cooltype command, since my changes speed up the cooldown system in general?

PR #1168, opened 2017-01-17
Filter resolved usernames to only valid characters

gmt2001:


BrandenB:
I don't really see a point in adding this, since if you're calling the function with a name that does not have valid characters its probably not a real user to begin with.
And since this will replace the unwanted characters it wont return the same username if it does not exist.

gmt2001:
It was intended as a concrete resolution to any form of touser injection

BrandenB:
That's what I thought. But I don't want to merge something that might break something else for that small issue. I'll just add a replace in js in case someone does try to post a link with the touser tag, then it wont come out as a clickable link.

Edit:
Also, if someone uses a display name that has non English char this pull-request would of broke that.

PR #1216, opened 2017-02-04
fix auction vulnerability

MegaGM:
Steps to reproduce: 
1. Start an auction
2. Somebody bids fairly
3. Someone sends in chat `!bid 3-0` or something else that can be successfully parsed by `parseInt()`
3. Regardless of anything, the person now is the `auction.topUser` and `auction.topPoints` now holds the parsed value from his message. `auction.topPoints === 3`

p.s. I'm glad to see you're using NodeBB, thank you for the choice

### DIFF
--- a/javascript-source/systems/auctionSystem.js
+++ b/javascript-source/systems/auctionSystem.js
@@ -116,7 +116,7 @@
             return;
         }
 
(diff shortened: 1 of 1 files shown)


BrandenB:
Good find! Thank you for this. <3

PR #1220, opened 2017-02-04
Separate dependencies from JDA; Fixed MSNeventListener

agaertner:
Two things:

JDA with dependencies bundles a dublicate version of commons-lang, outdated version of JNA.

**Separated** JDA-WithDependencies-*.jar from its dependencies:
**Added** all the same versions of libraries needed for JDA.
**Updated** commons-lang 3 3.3.2 to 3 3.5 and JNA to 4.3.0 (In the JDA-WithDependencies.jar bundle it is 4.2.2 which is missing WNDCLASSEX and other features)

Added my fixed version of MSNeventListener, which - at this moment - only supports microsoft windows meaning it starts after the SystemUtils.IS_OS_WINDOWS check.

Tested and works without any stacktraces and doesn't break anything.

Test output (Twitch chat):

```
[00:09] @^* TaokakaBot is listening to Agressor Bunx - Tornado (Original Mix)
[00:09] @^* TaokakaBot is listening to Natsume Chiaki - Hanairo Biyori
[00:09] @^* TaokakaBot is listening to Shindehai - ONE-PUNCH MAN FOREVER
[00:09] @^* TaokakaBot is playing Shindehai - ONE-PUNCH MAN FOREVER [Lumael's A-Class]
```
In the rhythm game osu you can write /np (for now playing) in the IRC and it displays in chat like the /me command in Twitch. I just made it look the same by using /me in `.say()`.

Since the rhythm game osu only really supports microsoft windows (OS X is still in early Alpha) this will only support windows.

For this to work you need to go into osu!'s options menu and turn on “MSN status integration” as this script is disguising a hidden native window as MSN to eavesdrop on the application message data sent from osu to the MSN messenger.


IllusionaryOne:
All features are to be compatible with all of the OSs that PhantomBot supports.  By restricting to only Windows, this goes against our charter.  I would suggest learning the OSU! API and making a compatible module for all OSs.  It seems they use IRC for their interface, if I remember right.

Also, the libraries are contained within that one Maven Distribution because they have been tested and certified together; by changing them, that introduces potential problems.  If there was a bug, we have no recourse with JDA at that point, because we would have indicate that we are not using the same versions of the API that they tested with.

Our apologies for this, but closing the request.

agaertner:
Even though they use IRC and they indeed have an API. There is no actual way of getting the current song except

- Window title changes (osu changes window title when playing, but not when listening in menus)
- eavesdropping on the MSN status integration (can be used when playing and in menus)

When I compiled my own version of PhantomBot with the same JDA version except for the dependencies layed out it did not break anything. I made sure to use the same dependency versions except for JNA and commons-lang.

Could you at least accept a pull request that would lay out the JDA dependencies so I (and others) have an easier time updating own compiled versions of PhantomBot?

Btw, since JDA-WithDependencies bundles commons-lang you have a dublicate in the libraries folder.

PR #1238, opened 2017-02-12
Fix index too large error in MySQLStore

gmt2001:
Remove unused imports


gmt2001:
Is there something wrong with this commit?

Hackebein:
I think it's better to display a warning/error if the databse has a wrong charset. I think this is the best solution for future test and support scenarios.

gmt2001:
UTF8MB4 is actually the correct charset to use in modern times.

BrandenB:
I'm going to merge this for now, I hope that it's fully tested.

PR #1266, opened 2017-02-24
Update slotMachine.js

nemmm:
Correct formula to calculate 30 percent and only from the pair (unlike the lowest value from all 3 emotes).


BrandenB:
Could you upload a screenshot of this working please?

nemmm:
image

(I'm using 20% for a pair in my script)

PR #1272, opened 2017-02-26
Check if 'message' is null in getGame

Invariel:
If the streamer was playing something, but the game name came back as 'null' (currently .twitch.tv/thespanishpresident), channelData.getString('message') will fail, throwing the error up to init.js and exiting out of the shoutout code.  This means that the user will not get the shoutout.

The proposed change checks for a null message, but results in followhandler.shoutout.no.game being printed to the channel, which is the intended result.


BrandenB:
Thanks!

PR #1277, opened 2017-03-01
Docker Support

Hackebein:
* Add Docker files
* Add Load up the bot info from the environment
* Move botlogin.txt for Docker support
* Clean up obsolet code

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -43,6 +43,7 @@ bin/
 # Files created by the bot inside resources folder
 resources/bannedUsers.bin
 resources/botlogin.txt
+resources/config/botlogin.txt
 resources/phantombot.db
 resources/stacktrace.txt
 resources/stdio.txt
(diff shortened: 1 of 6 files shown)


IllusionaryOne:
It looks like you have moved botlogin.txt, which impacts all of the current users.  That also requires updates to videos, documents, and forum posts.  I would ask to leave the configuration where it is if possible.  Or, sell us on why it needs to move.  We have discussed internally moving it, but, have decided not to in the past.

It also looks like you are relying on Docker Environmental variables - I am going to presume you have tested this on Docker and can show the bot properly functioning?  Also, a regression test in a non-Docker environment?

Hackebein:
Updates on videos and documentations is a good point. I had not thought about it.
It's possible to link a file, but it's recommended to use a folder. I revert it to the old position or add an optional parameter to define an other config position. The Envoriment updates the config file in the container. The config file must not be linked outside. I add the support only for users they prefer to update a file instead of docker container enviroment.

I test it with and without Docker. The Environment variable must begin with "PHANTOMBOT_".

Hackebein:
@IllusionaryOne can you review the code again?

Hackebein:
closed,  because i reforked the repo.

PR #1312, opened 2017-03-15
Fixed broken accountLink stuff

gmt2001:
accountLink and the example discord pointsSystem script were completely broken by some attempted changes by others before it was merged into master

### DIFF
--- a/javascript-source/discord/core/accountLink.js
+++ b/javascript-source/discord/core/accountLink.js
@@ -5,25 +5,41 @@
 (function () {
 
     /**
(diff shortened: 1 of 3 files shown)


BrandenB:
Not sure why you're saying it's completely broken, it's working fine for PhantomIndex and I. This pull-request reverts changes I made after merging it. 

These changes will also break all the other Discord systems that use the account link feature since you reverted the function `resolveTwitchName` back to `discordToTwitch`.

PR #1313, opened 2017-03-15
Added debugging for TwitchAPIv3 web responses

gmt2001:


BrandenB:
Any reason for adding this?

IllusionaryOne:
Finally, for a while we did write out the JSON errors to the Console.  This resulted in confusion from users who were not clear on what was being displayed.  As a professional software engineer for over 20 years, I can tell you that most users want to see an English description of an issue, not a "core dump" (this should show my age).  When you provide a user with 20 lines of JSON, that is meaningless, unless they are able to read JSON.  Projects that I work on, we aren't even allowed to dump such information to a user, and, in the end, the users of PhantomBot are primarily users that are not developers that understand things like JSON.

If anything, I could see this being written to a log file but not the Console.  Thoughts on that?

gmt2001:
I put it in debug mode only for a reason

IllusionaryOne:
Yes, and we tell our end users to put the bot into debug mode as well and we want to see short and concise data presented to help us isolate an issue.  This is why we have log files for more detailed data.  This is why we worked to remove stacktraces from going to the Console as well.  At my job, we use SOAP transports and never write the SOAP to the console, only to logs, why?  A SOAP request that is pages long serves no purpose flying by on a Console.  It needs to be placed in a file for research.  

Just write the data to the log file and use 'tail -f' and you have the same thing as a Console without spamming the Console with JSON.

gmt2001:
I can always patch it to log when I get home, or whoever merges can.

I think alot of the core needs some cleanup or restructuring, mostly the stuff from before you joined. That includes my stuff

PR #1320, opened 2017-03-16
DataStore Cleanup

gmt2001:
General code style cleanup for the DataStore classes
Moved DataStore classes to their own sub-package

### DIFF
--- a/javascript-source/discord/core/accountLink.js
+++ b/javascript-source/discord/core/accountLink.js
@@ -58,7 +58,7 @@
                     text += code.charAt(Math.floor(Math.random() * code.length));
                 }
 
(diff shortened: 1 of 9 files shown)


BrandenB:
I wouldn't mind merging this since it does cleanup imports and other things, but after talking to the other guys some of them don't like using uppercase for final values, and it wouldn't match the other classes. If you change it back to what it was I'll merge this, same thing goes for the other pull-request you created.

gmt2001:
I did it because, technically, the final values are constants and standard coding practice is to make constants all caps. I can undo it though if you want

BrandenB:
Yeah that I know, but it hasn't been done in any other classes, I don't really care either way but others do. If you have time to revert back do it and then I'll merge this.

gmt2001:
Done

PR #1323, opened 2017-03-17
Added names to Threads, where possible

gmt2001:


IllusionaryOne:
Downloaded locally, tested in jvisualvm and netbeans.

PR #1328, opened 2017-03-19
Updated deathcounter module

lbarnett83:
Unless I've missed something, the getArguments() was unnessecary.  Tested and working


BrandenB:
Thanks!

PR #1329, opened 2017-03-19
Dashboard Video Preview Addition

Pottymouth222:
All of the changes are written [here](https://community.phantombot.tv/topic/1905/dashboard-video-preview-modification).

(Sorry if I did something wrong with this pull, first time using github)

### DIFF
--- a/resources/web/panel/css/style.css
+++ b/resources/web/panel/css/style.css
@@ -342,13 +342,35 @@ td {
 #chatsidebar {
     position: absolute;
     top: 82px;
-    right: 10px;
+    right: 2px;
     bottom: 0;
     height: 650px;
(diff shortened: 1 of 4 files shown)


BrandenB:
Not sure about this, I'll talk to the team about it first.

BrandenB:
I've been told that the Twitch player uses a decent amount of memory, meaning adding it to the panel will increase its memory usage, and it may cause lag or crashes for users with larger channels. Since we do not want this to happen, we won't be merging this into our main build. We will look more into this feature for our new panel though!

Pottymouth222:
Alrighty, let me know if anything comes of it?  :+1:

PR #1339, opened 2017-03-26
Remove unnecessary Exception catches

gmt2001:
Condensed unnecessary specific Exception catches down to a single catch-all
Added the line number that generated the exception to the exception name

### DIFF
--- a/source/com/gmt2001/TwitchAPIv3.java
+++ b/source/com/gmt2001/TwitchAPIv3.java
@@ -135,7 +135,7 @@ public class TwitchAPIv3 {
             fillJSONObject(j, true, type.name(), post, url, c.getResponseCode(), "", "", content);
         } catch (Exception ex) {
             Throwable rootCause = ex;
-            while(rootCause.getCause() != null &&  rootCause.getCause() != rootCause) {
+            while (rootCause.getCause() != null && rootCause.getCause() != rootCause) {
                 rootCause = rootCause.getCause();
             }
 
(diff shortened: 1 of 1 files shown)


BrandenB:
Going to wait for IllusionaryOne to check this out since he updated it not long ago.

IllusionaryOne:
We would never allow this at any of the places that I have worked, this would fail most code reviews.  The exceptions that are thrown by each method should be known and caught, in case of any special handling needed.  At least, that is good coding style.  In this instance, I know that we do not provide any specific or special handling though for any of the caught exceptions.

I do see the benefit of condensing the code though, so will not provide a firm 'no' since we do not provide specific handling based on the exception, but, if it is ever warranted, it might be better to have it still split out to the specific exceptions.

gmt2001:
If you need special handling then add that back. I'm improving the output here because some sub-class exceptions are being caught as their parents thus reducing the amount of useful data we get from a report. This catch-all retrieves the real class name

IllusionaryOne:
How about just remove @SuppressWarnings("UseSpecificCatch") and catch all applicable exceptions?  Then the problem is solved, you have to catch all exceptions at that point and will not roll up an exception to a parent catch.

IllusionaryOne:
Anyway, will merge, if the following is fixed:
+            while(rootCause.getCause() != null &&  rootCause.getCause() != rootCause) {
to:
+            while (rootCause.getCause() != null && rootCause.getCause() != rootCause) {

Notice the two spaces before "rootCause.getCause()" and we typically put a space after 'while' statements.  Minor items, but, just do that and will merge.

gmt2001:
Done

IllusionaryOne:
Thanks, and thanks for having a healthy conversion over the changes.

PR #1349, opened 2017-04-04
Improved console based on Lanterna

gmt2001:
This is a new console based on [Lanterna](https://github.com/mabe02/lanterna). Lanterna is basically a curses-like library, written in pure Java (with optional native components), but with a little more functionality.

The primary advantages of this new console are:
 1. Cross-platform UTF-8 support (Windows users can finally see non-ASCII characters in their console!)
 2. A separate input box, so its actually possible to see what you type into the console when lots of stuff is going on (and use backspace effectively!)
 3. It is a stepping stone to potential cross-platform installer/updaters, or the ability to have certain important functions (Like the first run wizard) execute in a GUI-like environment, rather than the command line

The new console is automatically run when the **interactive** command line option is present (all non-service launcher scripts, at present). You can launch the bot with the native console (old way) by adding the **oldconsole** command line option (ie. add **-Doldconsole** after **java**)

Old calls to System.in and System.out have been replaced by selector methods in com.gmt2001.Console.in and com.gmt2001.Console.out which forward method calls appropriately, depending on whether the Lanterna GUI is being used or not. To the methods making the calls, it should be as if nothing changed.

Pasting is supported by using CTRL+V

NOTE: Due to the limitations of .bat files on Windows, the command prompt window will be open while the program runs. In its current form, the window must stay open for the bot to run. The java command can be changed to **javaw** to allow closing the window while the bot runs. There are also ways to fix it so that the command prompt will automatically disappear once the bot launches


BrandenB:
Here are things I noticed:

1. As you mentioned in your comment, using this will keep two consoles opened, for a streamer having more windows opened is never fun, a lot of people even ask on the forum if they can just run the bot in the background, which is possible, but it doesn't mean we should add a bunch of windows.

2. It uses a lot of memory, in my channel with the current repo build the bot will use about 40-50mb of memory, with this console it goes all the way over 150+mb, in a lager channel it's probably worse, we like trying to keep the bot from using a crap load of resources. Also, when compiling the bot with ant, it throws a few warnings about heap pollution, which I can clearly see in the netbeans profiler.

3. Higher CPU usage, usually when the bot is idle it will be hanging at 0% usage, with this console it's always around 2% which isn't a lot, but when you're running OBS and other streaming things on your computer it adds up, and again, this is without a load.

4. I just ran the bot in an event channel for a few minutes for fun, after a bit the console started to lag a bit, memory usage also got near 1gb.

5. Can't scroll, now i'm sure this might be a setting, but sometimes a scroll bar would appear and go away after a few seconds.

6. Errors printed with `com.gmt2001.Console.err` are printed in the other console.


That being said, not sure if this will get merged, it has a few issues and it is a massive lib for just a console that most people probably don't even look at. I'll talk about it more with IllusionaryOne when he comes back.

gmt2001:
The 2 consoles thing is only for Windows. We can change the launch script to fix it.

Scrolling is supposed to be disabled at the moment, I pushed a patch to fix a bug that was making it appear.

Will look into the error output.

The rest I kinda expected for now. I could swap the launch parameters around so you have to explicitly declare a launch parameter to use it (and let the native consoles be the default). When it is switched back to the native consoles there should be negligible, if any, impact on performance (just an if statement)

gmt2001:
The heap pollution error is referring to a use of varargs in a parameter which could allow assigning a wrong value type to the parameter if one is not careful, which then leads to a ClassCastException. It is not a problem if one uses the parameters correctly. I will suppress the warnings. http://stackoverflow.com/questions/12462079/potential-heap-pollution-via-varargs-parameter

gmt2001:
I just pushed an update to make lanterna non-default and added new launch scripts that enable it, using the **-Dnewconsole** option. I'm going to push a change in just a minute to fix the windows one to not keep the second window

BrandenB:
Still not sure with this, adding more files in the main folder makes it complicated for users who have no idea what they do, it also clusters the folder. Also like I mentioned before, it is a lot of code just for this console which doesn't really add features that will be useful to a normal user.

We'll wait see what IllusionaryOne has to say about it.

BrandenB:
You didn't have to close it, I said I was waiting on IO's thoughts about it, he's currently taking a small break. 

Sorry if I made you upset.

PR #1362, opened 2017-04-14
Update last follower if follower announcements are disabled

gmt2001:
Makes the streamInfo table (web panel latest follower) and latestFollower.txt file update on a new follow if followHandler is enabled, but followToggle (announce to chat) is false.

Also allows giving points to new followers in the same situation.


BrandenB:
This is more of a custom thing, replied here about it: https://community.phantombot.tv/topic/2060/recent-follower-not-updating-webpanel/6

PR #1389, opened 2017-04-28
Bug fixes and improvements

gmt2001:
Changed ScriptFileWatcher to a Singleton that scripts register with
ScriptFileWatcher is now stopped during PhantomBot::onExit
Moved ScriptFileWatcher registration to Script::load
Fixed lang scripts not being registered to ScriptFileWatcher on Windows when PhantomBot.reloadScripts was false
Changed some indexOf != -1 instances to contains()
Fixed PhantomBot::handleCommand(String, String, Channel) causing ScriptEventManager::runDirect to fail with a NullPointerException


IllusionaryOne:
Prefer the use of individual threads for all files.  10 seconds is too long to wait for updates.  

Implemented the lang issue differently in Build 3390.  Using the Java system routine to determine the OS file separator rather than performing two checks.

contains() at times is documented to be slower than indexOf().  In most Java implementations contains() simply calls indexOf().  

The null issue is handled at the root level object in another patch.

PR #1401, opened 2017-05-01
Portability Items

IllusionaryOne:
**launch-service.osx.sh, launch.osx.sh**
- Removed.  Code to detect macOS moved into launch.sh and launch-service.sh**

**launch-service.sh, launch.sh**
- Detects macOS and performs correct path handling
- No longer checks Java version

**PhantomBot.java**
- Detects Java version, exits if not new enough
- Prints out information about current Java version and OS

**Script.java**
- Portable code added for handling lang files


IllusionaryOne:
Tested scripts under macOS Sierra and Linux.  Tested Java version detection by setting the minimum version to 1.9 to ensure it failed, then back to 1.8 for production use.  Tested lang file update in Windows 10.

PR #1417, opened 2017-05-04
Delete dist/build and resources/scripts when running clean target

gmt2001:


IllusionaryOne:
distclean cleans up the dist/ directory.  The clean directive should not touch the dist hierarchy.

gmt2001:
Technically, that is a Build folder that is deleted in this patch, which does not belong in the dist folder either.

PR #1437, opened 2017-05-13
DataStore2

gmt2001:
Okay, this is ready for PR.

This is the DataStore2. It is intended to be used EXACTLY as I have written it. If you see something missing that was in original DataStore, it is by design. All members are fully javadoced (some by inheritance), but feel free to ask if you need clarification.

This is written as a complete replacement for DataStore in scripts. The intention is to transition us to real tables, designed and referenced properly, at a later date (especially when the bot is upgraded to TwitchAPIv5).

An example script, useridcollector.js, is included. It will collect TwitchAPIv5 user-id's from IRCv3 tags when someone speaks in channel as a precursor for the eventual transition. This script shows the proper way to create a table and execute a transaction using DataStore2, although not everything is shown.

DataStore will remain as a legacy interface for old scripts, scripts not yet converted, and programmers who don't know SQL. This is why I strongly urge, as in my example script, using *phantombot2_* as the prefix for properly designed future tables. If everything is working as it should, the only noticeable difference between the old implementation and the new implementation of legacy DataStore should be that you can not have a SqliteStore and a MySQLStore exist at the same time (must restart bot to switch, I reworked sqlite2Mysql to accommodate this), and that you can not use setAutoCommit to run a transaction on either one anymore (I really tried to make it work. SetBatchString works as a transaction though).

I have run tests on my own and things seem to be okay, but I would like some testing done on real setups with both SQLite and MySQL if possible before the PR is merged (please backup just in case). A full breakdown of the new package for the release notes will come at a later date, likely when conversions start.

### DIFF
--- a/source/com/gmt2001/HttpResponse.java
+++ b/source/com/gmt2001/HttpResponse.java
@@ -24,7 +24,7 @@ import java.util.HashMap;
  * @author gmt2001
  */
 public class HttpResponse {
-
+ 
     public RequestType type;
     public String url;
     public String post;


BrandenB:
Okay, this isn't the final review, I will talk with IllusionaryOne once he comes online later tonight. These are just things IllusionaryOne and I already agreed on when you mentioned you were making this in Discord.


- 1 The current scripts will not be refactored to work with this, there's too much of a risk of creating a bunch of new bugs, or corrupting user data. After the update from v1.6.3 to v2.0, it has taken us a very long time to get most of the scripts stable, and to this day there are still many bugs and issues.

- 2 I would like you to fully test this and prove that it all works, not just assume it's all good, we no longer test pull-requests because it just takes too much time, and it should be tested by the user who is submitting it, in our opinions.

- 3 The user id collector script is not needed, IllusionaryOne and I already agreed it was going to be placed in the `TwitchWSIRCParser.java` once all the events are done being sent. There's no point in storing all the ids in memory for this, it will just increase the usage for large channels.


This is all I have to say for now, IllusionaryOne and I will talk about this later, if he is on, and we'll decide if it should get merged or not.

gmt2001:
I did run tests on it. The reason I ask for someone else's help making absolutely sure it's good it's because I don't moderate/have access to any channels where I can run this and see it actually processing commands from lots of users

PR #1482, opened 2017-05-31
RepoVersion getting from Manifest.mf info

stachu540:
`RepoVersion.java` - for many attempts and test's this class now getting information about package from the MANIFEST file.
Also... there is changes about nightly build states. Now it is a boolean state. **Default is false**.

Ill leave this for reviewing and testing it.

### DIFF
--- a/build.xml
+++ b/build.xml
@@ -35,6 +35,18 @@
         <and>
             <isset property="nightly" />
             <length string="${nightly}" trim="yes" length="0" when="greater" />
+            <equals arg1="${nightly}" arg2="nightly_build" />
+        </and>
+    </condition>
+    <condition property="nightly" value="not_nightly">
+        <and>
+            <not>
+                <and>
+                    <isset property="nightly" />
+                    <equals arg1="${nightly}" arg2="nightly_build" />
+                </and>
+            </not>
+            <isfalse value="${nightly_build}" />
         </and>
     </condition>
     
@@ -72,7 +84,7 @@
     </target>
 
     <target depends="init,git.revision" name="pre.compile">
-        <echo>${ant.project.name} Version ${version} (${repository.version}) (${nightly_build})</echo>
+        <echo>${ant.project.name} Version ${version} (${repository.version}) (${nightly})</echo>
         <echo>${ant.project.name} Control Panel Version ${webpanel.version}</echo>
 
         <replaceregexp
--- a/build.xml
+++ b/build.xml
@@ -31,7 +31,7 @@
         </fileset>
     </path>
 
-    <condition property="nightly_build" value="${nightly}" else="not_nightly">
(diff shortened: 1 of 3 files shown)


IllusionaryOne:
Issues:

1) pre.compile:
     [echo] PhantomBot Version 2.3.6.1 (49f9a54) (true)
     [echo] PhantomBot Control Panel Version 1.1

That shouldn't say (true) I have tools that depend on seeing that is nightly on my build server.  Generally, do not change text in a build process.

2) /* astyle --style=java --indent=spaces=4 */

Didn't check beyond that.

stachu540:
`RepoVersion.java` - Code reformatted into the 4 spaces. (I love working on 2)
`build.xml` - added another one condition - now is not boolean word. It will be `nightly_build` or `not_nightly`.

Also i work on maven script for this repo. I need some coppule test's and changes. If you are interest change compiler I work for this stuff [here](https://github.com/stachu540/PhantomBot_Maven).

IllusionaryOne:
Will review later today.  Would not go with Maven.  For one:

    <dependency>
      <groupId>org.java-websocket</groupId>
      <artifactId>java-websocket</artifactId>
      <version>1.3.3</version> 
    </dependency>

I have had to make modifications and we use a specialized version of this library, the latest version there has even worse problems when I tested it.  There are bugs with SSL handling.  

Same goes for other directives in there, going to versions that we have tested out and have problems on a large scale would not be desired.

Good work though!

stachu540:
Well, I don't know mate 🤷‍♂️ . You are developers. It is up to you. I just getting fun for doing this, and getting some knowledge about it. 😄 

By the way. Merge is ready to other test's and pulls. For now... it is looks good, if you ask for it.

Very important thing❗️ If try some changes on Manifest file, it will be affect whole project (maybe not not all). but only in `jar` file. Be sure, if everything is fine. Otherwise, I'll close this pull request. I have too much work for other projects.

Thanks for the advice IO. 👍

BrandenB:
I'm with IllusionaryOne with Maven, some of the libraries we use are probably really outdated, some are modified, meaning if we update them we will have to change some of the bot's code.

I will merge this change for now though.

BrandenB:
I will be reverting these changes since it seems to not work for some users.

More information is available here: https://community.phantombot.tv/t/phantombot-2-3-7-1-update-issue/327/11

stachu540:
If Manifest doesn't exists, it will throw exception and returning null. I don't get it why he is return null and doesn't throw `NullPointException`?

I'll investigate it, for those reason. By the way, **reverting changes is a final option**.

IllusionaryOne:
I reverted the change last night, and that worked for the user.  Still not entirely sure why this is the case, the code seems solid, but, for now, as the previous way is essentially foolproof, going with that for right now.

PR #1507, opened 2017-06-14
Discord Host Handler Fix

IllusionaryOne:
**hostHandler.js**
- Discord did not like viewers being an integer, had to be casted as a
String object when placing in the output string.


IllusionaryOne:
This grabbed other commits it should not have.

PR #1522, opened 2017-06-17
fix wrong method call, optimized deathcounter check

3stadt:
There was a call to `$.deathUpdateFilegame();` in the deathcounter command.

Additionally, I pulled up the deathcounter var that was declared twice.

I also intruduced a new statement that declares a var holding the result of the deathcounter existance check and at the same time sets the deathCounter var to 0 if it's 0 or NaN.


BrandenB:
Thank you, merging this!

PR #1528, opened 2017-06-18
Gather output in the new logfile PhantomBot.out

shiek:
No more output to /var/log/syslog when using systemd

### DIFF
--- a/resources/launch-service.sh
+++ b/resources/launch-service.sh
@@ -30,4 +30,4 @@ else
     echo "You don't have Java installed! Download it from https://www.java.com/en/download/"
 fi
 
-java -Dfile.encoding=UTF-8 -jar PhantomBot.jar
+java -Dfile.encoding=UTF-8 -jar PhantomBot.jar >>PhantomBot.out 2>&1


IllusionaryOne:
I personally use journalctl to watch the logs and allow the built in syslog rotation to handle rolling off old logs and not worrying about a local file growing too large.  If it does grow too large, then folks will need to know to "echo" to the file and not remove it.  For example, under macOS, if you remove the file, the handle is lost and no output is written at that point.  Same in Linux when I double checked there.  I remembered this from working in Solaris and we always used logrotate for this reason for any application.

This could confuse people that do not know to handle the logfile a certain way, I think.

Thoughts?

shiek:
I really don't like my syslog spammed with bot stuff.
Especially in the launch-service.sh case bot stuff should stay in the bot folder.
Unless somebody has a crazy big channel logfile size won't be an issue.
The size of the PhantomBot.out should be 10 MB per year for a normal channel.

IllusionaryOne:
If the file is removed the space is not reclaimed until the application shuts down, which can confuse folks.  10MB doesn't sound right either, on my VPS without debugging enabled, when I check the size of the journal log, without debugging enabled and without streaming to my channel:

journalctl --no-pager | grep launch-service | wc -c
703321496

...or 703MB.   If I was redirecting to disk and deleted the file, that space is not reclaimed until I stop the bot process.  In other words, in Linux (and I would need to check macOS) - if you delete a file that is being redirected to, data is still written to the device regardless and fills up space, but you will never find the file.  Someone may wonder why their tiny instance on AWS gets filled up if they delete a file, and the free space is not reclaimed.  To illustrate, when the bot running:

l-wx------ 1 illusion illusion 64 Jun 19 17:38 1 -> /usr/local/opt/iobot2/file

File removed:

l-wx------ 1 illusion illusion 64 Jun 19 17:38 1 -> /usr/local/opt/iobot2/file (deleted)

The handle stays open after the file is removed.

Unfortunately, I have to close the request.  I would say keep a local copy of the script that acts as you want.

PR #1531, opened 2017-06-19
Consolidate Check Port Logic

IllusionaryOne:
**PhantomBot.java**
- Was calling the same checkPortAvailability() code twice for SSL and non-SSL server creation - now just call once.
- Added baseport to the list of ports to check.


IllusionaryOne:
Merging as low impact change.  Just moved around the calls to methods.

PR #1562, opened 2017-07-11
Fix clearing the timer after close a bet and fixed bet close with no args not working properly

nemmm:
1. timeout variable was not being used and it was needed to prevent bugs related to the timer.
2. When you use join there is not possible undefined option variable, there will have blank content or blank space


nemmm:
Sorry I'm noob on github..

BrandenB:
Using `args[0]` only allows for one word where there could be multiple on some options. Will be pushing a fix, thanks for this.

PR #1585, opened 2017-07-30
Make the Current Ranks css match the same style input boxes

CreativeCodeCat:
just adding css to the Current Ranks so they match the style of making ranks input boxes

### DIFF
--- a/resources/web/panel/css/style.css
+++ b/resources/web/panel/css/style.css
@@ -808,3 +808,26 @@ input, textarea {
     border-color: grey;
 }
 
+
+
+
+
+
+.rank-control {
+    display: block;
+    width: 100%;
+    height: 34px;
+    padding: 6px 12px;
+    font-size: 14px;
+    line-height: 1.42857143;
+    color: #555;
+    background-color: #fff;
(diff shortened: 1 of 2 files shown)


CreativeCodeCat:
Below is a list of other ones that need the css to look nicer too

Current Command Prices
Current Command Payments
Commands with Cooldowns


these are a few i have tested but it seems like any that adds new input's below the settings

resources/web/panel/css/style.css
```diff
@@ -808,3 +808,26 @@ input, textarea {
     border-color: grey;
 }
 
+
```
BrandenB, line 811: Can you remove all of these extra spaces please?

BrandenB:
Can you provide a screenshot please?

resources/web/panel/css/style.css
CreativeCodeCat: i wasn't sure where in the css it would go so i just stuck it there for you guys to make sure it goes in a neat spot :)

CreativeCodeCat:
Closing this as I will put it in my next pull

PR #1587, opened 2017-07-31
Allow fast editing of most of the colors

CreativeCodeCat:
here is a push of most of the panel for custom colors using the root: there may be more needed but this is what worked for me

### DIFF
--- a/resources/web/panel/css/style.css
+++ b/resources/web/panel/css/style.css
@@ -1,3 +1,17 @@
+:root {
+    --header-background: #6441a5;
+    --main-background: #201c2b;
+    --ui-widget-content-border: #6136b1;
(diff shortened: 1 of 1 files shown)


CreativeCodeCat:
http://pussh.us/img/1501528714_877.gif

Screenshot of what this helps do!

CreativeCodeCat:
oops i forgot to change 
```
#chatsidebar-header {
    margin: 0;
    text-align: right;
    padding-right: 5px;
    background: rgb(36, 36, 36);
    color: rgb(100, 66, 166);
}
```

To

```
#chatsidebar-header {
    margin: 0;
    text-align: right;
    padding-right: 5px;
    background: rgb(36, 36, 36);
    color: var(--ui-widget-content-background);
}
```

PR #1598, opened 2017-08-04
Updated Dockerfile to use multi-stage build

m3adow:
This utilizes the new Docker multi-stage build feature to push image size down from 982MB to 134MB. 

Disadvantage: Only available from Docker 17.05 upwards.


BrandenB:
Thanks!

PR #1603, opened 2017-08-05
Fixed CSS - Notices & Settings also made adventure top5 as a sub command so it can have its own permissions

CreativeCodeCat:
something forgotten but now fixed


CreativeCodeCat (commented):
. not #

BrandenB:
I would merge the fixes for the panel, but not the custom TTS script, sorry.

PR #1606, opened 2017-08-07
Fixed the css and added sub command

CreativeCodeCat:
fixed the other css and added the sub command fior adventure so u can set the top5 to a different permission


IllusionaryOne:
After further review, the stock colors were changed.  Considered backing out but fixing the colors instead - next time we will reject a PR that changes our stock color.  Those are part of our brand.

CreativeCodeCat:
https://github.com/PhantomBot/PhantomBot/blob/984d91de177c46ce3d5846afd1cbae5951ee5b7e/resources/web/panel/css/style.css

http://pussh.us/img/1502226127_e63.png

.ui-tabs-vertical .ui-tabs-nav .ui-state-hover, .ui-state-focus was 6136b1 not 6441a5 which is what u you set it too now

PR #1613, opened 2017-08-11
Fixed typos (moderation.html)

PowerJax:


BrandenB:
Good catch! Sorry for the delayed response.

PR #1622, opened 2017-08-17
Added quoteSystem to Discord

CreativeCodeCat:
i have added the quoteSystem to Discord


BrandenB:
Sorry for the late reply, I would consider merging this if Discord quotes had their own table in the database, also, this module would have to be added to the panel. It would also have to be disabled by default.

BrandenB:
Closing this for now.

PR #1640, opened 2017-08-25
Moved jquery files to common directory

HighHowDoIPlay:
Removed some extra weight from the web files by combining like-files and moving them to the common folder. Everything looked the same on my end, so I hope there's no issues.


BrandenB:
Looks good on my end too, it also removes a bunch of files. Thank you!

PR #1643, opened 2017-08-26
Added Auto Game Change to the Tweet game change!

CreativeCodeCat:


CreativeCodeCat:
Nvm this is useless cus you have to change games to trigger it

PR #1665, opened 2017-09-13
Update core-accountLink.js

Tingyoma:
I have added a missing part to the discord account link required string


BrandenB:
This was intended, just using `!account` gives the user another response about how the system works if they ever change their Twitch name.

PR #1687, opened 2017-09-27
Fixed successful discord account linking message

HighHowDoIPlay:
Fixed the successful discord account link message from stating the senders twitch url to the channel owner's instead.

Also fixed a comment typo because why not?


BrandenB:
Thanks!

PR #1703, opened 2017-10-04
Fix setconnectmessage command from including itsetlf in message

HighHowDoIPlay:
This fixes the !botname setconnectmessage command from including 'setconnectmessage' in the message.

action was changed to subAction since action is already being checked in the main if-statement, and subAction will ensure there's a word after setconnectmessage.


BrandenB:
Thanks.

PR #1704, opened 2017-10-06
Fix various bugs

Caricatore:
**systems/ticketraffleSystem.js**
- Ticket raffle entries counter was counting the same users, that had already bought tickets earlier.

**init.js**
- Missing an argument when checking subcommands permission in discord.

**discord/commands/customCommands.js**
- Incorrect work of function, when trying to add a list of channels through the !channelcom command.

**discord/games/gambling.js**
- “.discord” repeats in that string twice.

**web/panel/js/pointsPanel.js**
- Impossible to set “0” value through the web panel.
- Removing spaces from nicknames, to prevent creating dublicates in points database.

**web/panel/js/discordPanel.js**
- Fix aliases.
- Fix adding channels.
- “Checked” option wasn’t sending proper status from the discord web panel.

### DIFF
--- a/javascript-source/commands/customCommands.js
+++ b/javascript-source/commands/customCommands.js
@@ -135,11 +135,13 @@
         }
 
         if (message.match(/\(1=[^)]+\)/g)) {
+            var t = message.match(/\(1=[^)]+\)/)[0];
             if (event.getArgs()[0]) {
-                var t = message.match(/\(1=[^)]+\)/)[0];
                 message = $.replace(message, t, event.getArgs()[0]);
+            } else { 
+                var tR = t.replace('(1=', '').replace(')', '');
+                message = $.replace(message, t, tR);
             }
-            message = $.replace(message, '(1=', '(');
         }
 
         if (message.match(/\(countdown=[^)]+\)/g)) {
--- a/javascript-source/commands/customCommands.js
+++ b/javascript-source/commands/customCommands.js
@@ -135,13 +135,11 @@
         }
 
         if (message.match(/\(1=[^)]+\)/g)) {
-            var t = message.match(/\(1=[^)]+\)/)[0];
             if (event.getArgs()[0]) {
+                var t = message.match(/\(1=[^)]+\)/)[0];
                 message = $.replace(message, t, event.getArgs()[0]);
-            } else { 
-                var tR = t.replace('(1=', '').replace(')', '');
-                message = $.replace(message, t, tR);
             }
+            message = $.replace(message, '(1=', '(');
         }
 
         if (message.match(/\(countdown=[^)]+\)/g)) {
--- a/javascript-source/core/misc.js
+++ b/javascript-source/core/misc.js
@@ -155,7 +155,7 @@
      * @param {boolean} run
      */
     function sayWithTimeout(message, run) {
-        if (((timeout + 10000) > systemTime()) || !run) {
+        if (((timeout + 10000) > $.systemTime()) || !run) {
             return;
         }
 
--- a/javascript-source/core/misc.js
+++ b/javascript-source/core/misc.js
@@ -155,7 +155,7 @@
      * @param {boolean} run
      */
     function sayWithTimeout(message, run) {
-        if (((timeout + 10000) > $.systemTime()) || !run) {
+        if (((timeout + 10000) > systemTime()) || !run) {
             return;
         }
 
--- a/javascript-source/discord/commands/customCommands.js
+++ b/javascript-source/discord/commands/customCommands.js
@@ -454,10 +454,18 @@
                 return;
             }
 
-            $.inidb.set('discordChannelcom', action, subAction);
             $.discord.clearChannelCommands(action);
-            $.discord.setCommandChannel(action, subAction, false);
-            $.discord.say(channel, $.discord.userPrefix(mention) + $.lang.get('discord.customcommands.channelcom.success', action, subAction));
+            var key = [];
+            for (var i in args) {
+                if (i != 0) {
+                    args[i] = args[i].replace(',', '').toLowerCase();
+                    $.discord.setCommandChannel(action, args[i], false);
+                    key.push(args[i]);
+                }
+            }
+            key = key.join(', ');
+            $.inidb.set('discordChannelcom', action, key);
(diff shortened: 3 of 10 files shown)


IllusionaryOne:
$.systemTime() is not required in misc.js as the method is defined locally, it is in scope.  The export is for other scripts.

Illustrate example with customCommands.js

Bug has been created with changes to Discord customCommands.js.

I will finish reviewing later.  Going to dinner :)

Caricatore:
@IllusionaryOne do you mean to illustrate example about brackets (1= ) in customCommands.js ?
There is "!kiss [optional]" command that doing "(playsound kiss) (senderrank) kisses (1=Mikerina) <3 bleedPurple <3 bleedPurple"
When you typing !kiss without an argument it sends a message that contains nickname (Mikerina) in brackets, but we don't need those brackets :) Only when you use it with option, !kiss IllusionaryOne, for example, there are no brackets for now.

javascript-source/core/misc.js
```diff
@@ -155,7 +155,7 @@
      * @param {boolean} run
      */
     function sayWithTimeout(message, run) {
-        if (((timeout + 10000) > systemTime()) || !run) {
+        if (((timeout + 10000) > $.systemTime()) || !run) {
```
BrandenB: `$.` isn't needed since this function is in the local script, no point in calling the global scope for it.

javascript-source/discord/commands/customCommands.js
```diff
@@ -454,10 +454,18 @@
                 return;
             }
 
-            $.inidb.set('discordChannelcom', action, subAction);
             $.discord.clearChannelCommands(action);
-            $.discord.setCommandChannel(action, subAction, false);
-            $.discord.say(channel, $.discord.userPrefix(mention) + $.lang.get('discord.customcommands.channelcom.success', action, subAction));
+            var key = [];
```
BrandenB: This was not made to be automated, it took me a while to get the logic down, i'm afraid what changing it might cause other issues later on.

javascript-source/handlers/subscribeHandler.js
```diff
@@ -57,7 +57,7 @@
             $.writeToFile(subscriber + ' ', './addons/subscribeHandler/latestSub.txt', false);
             $.inidb.set('streamInfo', 'lastSub', subscriber);
             if (subReward > 0) {
-                $.inidb.incr('points', subscriber, subReward);
+                $.inidb.incr('points', subscriber.toLowerCase(), subReward);
```
BrandenB: Twitch sends this name in lowercase already, no need to do it twice.

javascript-source/handlers/subscribeHandler.js
```diff
@@ -82,7 +82,7 @@
...
```
BrandenB: Twitch sends this name in lowercase already, no need to do it twice.

javascript-source/handlers/subscribeHandler.js
```diff
@@ -120,7 +120,7 @@
             $.writeToFile(resubscriber + ': ' + months + ' ', './addons/subscribeHandler/latestResub&Months.txt', false);
             $.inidb.set('streamInfo', 'lastReSub', resubscriber);
             if (reSubReward > 0) {
-                $.inidb.incr('points', resubscriber, reSubReward);
+                $.inidb.incr('points', resubscriber.toLowerCase(), reSubReward);
```
BrandenB: Twitch sends this name in lowercase already, no need to do it twice.

javascript-source/commands/customCommands.js
```diff
@@ -135,11 +135,13 @@
         }
 
         if (message.match(/\(1=[^)]+\)/g)) {
+            var t = message.match(/\(1=[^)]+\)/)[0];
```
BrandenB: This function is made to accept command variables, not random text. It it working fine.

BrandenB (changes requested):
If you make the changes requested, we will merge it as the rest of this seems fine to me.

javascript-source/commands/customCommands.js
Caricatore: But if I understand correctly, there is (1) only for variables, and (1=default) here, if you want to see something by default to be written by the bot in chat, without brackets. As when you use variables, there are no brackets around them in text message.

javascript-source/discord/commands/customCommands.js
Caricatore: Without this changes, it's currently impossible to !channelcom command to more than one channel. It can be global or used only in one text channel. I don't know about future changes, but if you think it may cause troubles, I can cancel this commit. But then it shouldn't be said "spectate the channels with a comma and space for multiple" here https://phantombot.tv/commands/discord/ in documentation.

javascript-source/commands/customCommands.js
BrandenB: There's a bunch of variables, this meant to be used like this. `(1=random)`, if the first argument isn't passed, it will use the `(random)` variable.

javascript-source/discord/commands/customCommands.js
BrandenB: The command works as documented, at least it did when I tested the last time.

javascript-source/commands/customCommands.js
Caricatore: Ah, I see now...

javascript-source/discord/commands/customCommands.js
Caricatore: It can't work properly because when you try to add several channels like this
!channelcom somecommand chan1, chan2, chan3
it adds only the first argument "subAction = args[1];", "chan1," (with comma) in DB $.inidb.set('discordChannelcom', action, subAction); and via discord function $.discord.setCommandChannel(action, subAction, false);

BrandenB: You need to use quotes, I forgot to mention that in the documentation. `!channelcom "chan1, chan2, chan3"`

Caricatore: So quotes do the trick then :) ok

BrandenB: We can simply join the args too, that's another possibility.

BrandenB (approved):

deitylink1:
Sorry, github noob - ignore my pull. I thought I was just downloading the source (like like perforce). Now I see the correct link...

PR #1705, opened 2017-10-06
Discord: Multiple aliases for a command

Caricatore:
It would be great if you could make more than one alias for commands in discord.
_Files to update:_

**discord/commands/customCommands.js**
- Swapping arguments [alias] and [command] in !aliascom command
- Making it possible to add several aliases (separated by comma and space) to a command
- Fix @event webPanelSocketUpdate

**discord/core/registerCommand.js**
- Updating functions that work with aliases

### DIFF
--- a/javascript-source/commands/customCommands.js
+++ b/javascript-source/commands/customCommands.js
@@ -135,11 +135,13 @@
         }
 
         if (message.match(/\(1=[^)]+\)/g)) {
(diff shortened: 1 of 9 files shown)


BrandenB:
Please don't create pull requests with new features and bug fixes, it makes it almost impossible to review. Also, aliases aren't made to be this complex, if they were, I would've made it like our Twitch system.

PR #1747, opened 2017-11-03
- Fixed skip button loop again

BrandenB:


resources/web/ytplayer/js/ytPlayer.js
```diff
@@ -3,7 +3,7 @@
  * ------------
  * Interface for the YouTube Player for PhantomBot.
  */
-var DEBUG_MODE = false;
+var DEBUG_MODE = true;
```
ddugovic, line 6: I wonder, is this change intentional?

BrandenB:
It was reverted.

PR #1753, opened 2017-11-11
Fix for discord's custom command module disabled message

HighHowDoIPlay:
Replaced "games" with "commands" to no longer confuse users on which module they need to enable.


kojitsari:
You have a few extremely old, unnecessary commits tacked onto this. Do you want to try to fix this so it can be merged or would you rather one of us pull the fix?

PR #1770, opened 2017-12-02
Fix the toggle on the Panel for the Gifted Sub Alert

IllusionaryOne:
**greetings.html**
- Was using incorrect DOM name

**greetingsPanel.js**
- Was using incorrect DOM name

### DIFF
--- a/resources/web/panel/greetings.html
+++ b/resources/web/panel/greetings.html
@@ -371,7 +371,7 @@
                 <td style="width: 25px"><div id="subscriptiongiftGreetings" /></td>
                 <td style="width: 25px">
                     <div data-toggle="tooltip" title="Toggle" class="button"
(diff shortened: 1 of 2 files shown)


IllusionaryOne:
Toggled from GUI, you can see when I toggled it in between in the Console that the value changed as the same setting is applied twice:
```
giftsubwelcometoggle
[12-01-2017 @ 21:47:30.790 MST] [CHAT] @IllusionaryBot, Subscriber gift will now be welcomed upon subscribing.
giftsubwelcometoggle
[12-01-2017 @ 21:47:35.245 MST] [CHAT] @IllusionaryBot, Subscriber gift will no longer be welcomed upon subscribing.
giftsubwelcometoggle
[12-01-2017 @ 21:48:40.146 MST] [CHAT] @IllusionaryBot, Subscriber gift will no longer be welcomed upon subscribing.
```

PR #1773, opened 2017-12-06
Fix Issue with RegEx on Keywords Panel

IllusionaryOne:
**keywordsPanel.js**
- Created cached lists for edit, cooldown, and prices for keywords.
- Reference these by index rather than attempting to use the keyword as an index.


IllusionaryOne:
Tests:

- Created regex: word\sword\sword\s

Was able to edit, delete; add as a cooldown, edit that time, delete; add as a price, edit that price, delete.

PR #1774, opened 2017-12-06
Fix $.users Object Misassignment

IllusionaryOne:
**permissions.js**
- The users object was being assigned to newUsers which was then losing the scope on $.users
- Switched to using the previously created updateUsersObject() method which properly "resets" the users object.

### DIFF
--- a/javascript-source/core/permissions.js
+++ b/javascript-source/core/permissions.js
@@ -641,6 +641,12 @@
             for (var i in usernames) {
                 newUsers.push(usernames[i]);
             }
+
+            // Sometimes TMI has not seen the bot yet, this can happen at startup.
+            if (!hasKey(newUsers, $.botName)) {
+                newUsers.push($.botName);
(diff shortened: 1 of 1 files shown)


IllusionaryOne:
To show the issue.  When I was checking $.users and users in permissions.js:
```
users
[12-06-2017 @ 13:44:27.010 MST] users   >>> scaniatv,1512593057230
[12-06-2017 @ 13:44:27.010 MST] $.users >>> illusionarybot,1512593053048
```
Once TMI was called, $.users was being disassociated from users.  Now, after TMI is called:

```
users
[12-06-2017 @ 14:22:46.719 MST] users   >>> illusionarybot,1512595218021,scaniatv,1512595220228
[12-06-2017 @ 14:22:46.719 MST] $.users >>> illusionarybot,1512595218021,scaniatv,1512595220228
```
Which is what I would expect.

PR #1789, opened 2017-12-29
Fix for !ytp blacklist add/remove not counting multiple args

HighHowDoIPlay:
!ytp blacklist only counted the first arg after "add" or "remove", sometimes users put strings with spaces in the blacklist. This fix allows them to remove them.

Before modification:
!ytp blacklist remove hollywood undead - gangsta sexy [lyrics

would reply with:
Song/Keyword: "hollywood" has been removed from the blacklist.

(without deletion)

After modification it replies with:
Song/Keyword: "hollywood undead - gangsta sexy [lyrics" has been removed from the blacklist.

And successfully deletes the entry.

Note that for some reason github included my pull to update my branch to the master.... Not exactly sure why it did that again...


BrandenB:
Can you post screenshots of before and after your fix? Thanks!

HighHowDoIPlay:
Here's a before image:
Before
And after:
After
If you need more let me know.

BrandenB:
Good enough, thanks!

PR #1834, opened 2018-01-21
Uniformize file paths in the dev env installation guide.

DarckCrystale:
Minor change not affecting the bot (only a .md file was edited).

### DIFF
--- a/DEVSETUP.md
+++ b/DEVSETUP.md
@@ -10,7 +10,7 @@ It's very global, so you might need to take some extra steps to get it to work f
 1. Fork the [PhantomBot repository](https://github.com/PhantomBot/PhantomBot).
 2. Check out your for and open it in your favorite IDE.
 2. Set the project Java SDK to [1.8](http://www.oracle.com/technetwork/java/javase/overview/index.html).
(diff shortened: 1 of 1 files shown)


CreativeCodeCat:
the ./ isnt needed as its telling you to get the file not the path

DarckCrystale:
@JustAlixe l26, you have `./development-resources/watchers.xml` with a `./` and it's also a file.

IllusionaryOne:
In the IDEs for this documentation, you would select `build.xml` and not `./build.xml`.  Technically, it is the same, but, in the GUI it is not presented with the leading `./`.

image

I cannot speak to the items that I did not add, but I believe I committed the original instructions for build.xml.  The adjustment probably should be on the other paths -- but, I cannot speak to the other tool sets, only Eclipse.

IllusionaryOne:
So, should the other `./` be removed? I believe that is the case. @ScaniaTV any thoughts? It is documentation, not code.

BrandenB:
Unless it's actually an issue with some IDEs, I don't see a point, no.  Up to you if you want to merge @IllusionaryOne.

IllusionaryOne:
I am fine with closing this out and going in and changing the documentation to not have any leading `./`.  Closing.

PR #1835, opened 2018-01-21
Made a link internal to avoid problems after a fork.

DarckCrystale:
Minor edit of the ReadMe.md file.


IllusionaryOne:
Rejected.  The link needs to be there so that if someone has checked out the file to their local environment they can always open the latest in GitHub.

PR #1844, opened 2018-01-25
Added Foorter timestamp

CreativeCodeCat:
Added timestamp to all the embeds for the going live, game change and going offline

Example
image goes here


BrandenB:
This looks good, gonna wait for @IllusionaryOne to check it out and merge if he's okay with it!

IllusionaryOne:
Approved.  Dumb question though, if someone knows the answer and can reply (I am still merging it).  What timezone does the time report in?  Is that a nice Discord API call so it uses the local time for the Discord user or is it something else?  I just ask in case a user comes back and asks us if it isn't what they expect.

CreativeCodeCat:
@IllusionaryOne i believe its the local time of the user viewing it and it changes from today to yesterday and then to Last Thursday or what ever day it was on and so on :)

PR #1855, opened 2018-02-02
Implemented Twitch UserID tracking and automatic name change handling

gmt2001:
Completes trello card https://trello.com/c/hl95xq7s

*** tv.phantombot.event.twitch ***
Added package tv.phantombot.event.twitch.user
Added class tv.phantombot.event.twitch.user.TwitchUserNameChanged with properties userID, oldName, and newName

*** UsernameCache ***
Added fields int intervalLookupMS, ScheduledThreadPoolExecutor tpExec, and Queue lookupQueue
Changed type of field cache to ConcurrentMap
Replaced deprecated call to Maps.newHashMap with diamond inference
Added overloads to method addUser
Replaced all calls to cache.put and lookupUserData with addUser
Added private method checkUserNameChanged
Changed addUser to add to the cache if all data is available
Changed addUser to enqueue usernames in lookupQueue if all data is not available

*** init.js ***
Added bind for twitchUserNameChanged

*** nameConverter.js ***
Moved name change execution logic to seperate function
Changed name change execution logic to use a transaction
Counted updating one or more quotes as +1 to number of changed tables
Updated !namechange to call the function
Added bind for twitchUserNameChanged

*** permissions.js ***
Added call to $.username.addUser to binds for ircChannelMessage, ircChannelJoin, and ircChannelUsersUpdate

*******************************

This update implements tracking of Twitch UserIDs. When a user enters the channel or gets resolved,
their information is added to the UsernameCache using IRCv3 tags, as before, if available.

If it is not available, the user is now queued to have their information looked up. This is
implemented as being done at a rate of 1 API call per second (if something is queued), to
prepare for helix rate limits. Each API call will attempt to retrieve up to 100 users.
The downside of this change is that if a user is resolved before they have been looked up, their
username will be returned instead of doing an instant lookup.

When addUser successfully detects valid data and adds the user to the cache, it calls a method
that looks up the UserID in the new userids table. If the UserID (table key) is found, it checks
the stored username (table value) against the username being cached. If they do not match, the
TwitchUserNameChangedEvent is posted to the EventBus. The stored username is inserted if it doesn't
exist or updated if it has changed.

init.js handles the event as a forced hook call, so that disabled scripts handling their own
name changes can update their tables in case they later get re-enabled.

The new overloads for addUser allow data to be inserted from existing data, or IRCv3 tags. addUser
now considers userID invalid, and therefore enqueues a user for lookup, if the userID is 0.

The changing of cache to a ConcurrentMap type is required for thread safety.

The use of a ScheduledThreadPoolExecutor is preferable to an infinite loop with Thread.sleep or a Timer.

*******************************

Here are my test screenshots showing the points table and the new userids table in the database, initialized to a before name change state.
https://1drv.ms/i/s!AqJWOcIz2x-Ugt1CW9iVoX-cqsRvDg
https://1drv.ms/i/s!AqJWOcIz2x-Ugt1DfgkFN16kcXi11A

These screenshots show the database state after starting the bot, joining chat, and saying hi (to fast-trigger the resolve).
https://1drv.ms/i/s!AqJWOcIz2x-Ugt1I6zxG3CqeZxFFAg
https://1drv.ms/i/s!AqJWOcIz2x-Ugt1JIAV9qUJvK-4zeA

And here is the database after running `!namechange gmt2001 gmt2001testing` to make sure the command still works. I left the command in for edge cases or for doing conversions while the user isn't there.
https://1drv.ms/i/s!AqJWOcIz2x-Ugt1NT5EOjCZdkhs0xw
https://1drv.ms/i/s!AqJWOcIz2x-UguJDBjNhBtg3v5YZ3A

Finally, here is a screenshot of the console, showing the bot isn't broken.
https://1drv.ms/i/s!AqJWOcIz2x-UguJRTwUN7Pi0rX7xWg

*******************************

This setup should be able to handle Helix when you later implement it. I chose my rate limit as half the authorized Helix rate limit. My proposed solution to Helix's two rate limits is to use the chat_login token if no apiOAuth token is provided to get the increased rate limit, but still disable the services that require an apiOAuth if it is not provided.

### DIFF
--- a/javascript-source/commands/nameConverter.js
+++ b/javascript-source/commands/nameConverter.js
@@ -1,4 +1,46 @@
 (function() {
+    
+    /*
+     * @function changeTwitchName
+     */
+    function changeTwitchName(oldName, newName) {
+        var tables = ['points', 'time', 'followed', 'subscribed', 'visited', 'group', 'panelchatuserstats', 'panelchatstats', 'gamewispsubs'],
+            changed = 0,
+            i;
+
+        $.inidb.setAutoCommit(false);
+
+        // Update the default tables with that users new name if it's currently in any tables.
+        for (i in tables) {
+            if ($.inidb.exists(tables[i], oldName.toLowerCase()) == true) {
+                $.inidb.set(tables[i], newName.toLowerCase(), $.inidb.get(tables[i], oldName.toLowerCase()));
+                $.inidb.del(tables[i], oldName.toLowerCase());
+                changed++;
+            }
+        }
+
+        // Update the username in the quotes table.
+        var keys = $.inidb.GetKeyList('quotes', ''),
+            jsonArr,
+            foundAQuote = false;
+
+        for (i in keys) {
+            if ($.inidb.get('quotes', keys[i]).toLowerCase().indexOf(oldName.toLowerCase()) > -1) {
+                jsonArr = JSON.parse($.inidb.get('quotes', keys[i]));
+                $.inidb.set('quotes', keys[i], JSON.stringify([String(newName), String(jsonArr[1]), String(jsonArr[2]), String(jsonArr[3])]));
+                foundAQuote = true;
+            }
+        }
+        
+        if (foundAQuote) {
+            changed++;
+        }
+        
+        $.inidb.setAutoCommit(true);
+        
+        return changed;
+    }
+    
     $.bind('command', function(event) {
         var sender = event.getSender(),
             command = event.getCommand(),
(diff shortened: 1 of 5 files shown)


BrandenB:
As seen on Trello, this was only going to be implemented if the new API (Helix) allows it with its rate limits. Currently with version 5, you can spam it as much as you like and no calls will be rejected. Helix will have different limits for each endpoints. Helix will be implemented by @IllusionaryOne and I sometime this summer.

Also, if this ever gets done with Helix, every user in the database would get changed to an ID, this PR is sort of the "hacky" way of doing it.

Until TMI gives us the user's ID, this won't be implemented.

PR #1856, opened 2018-02-02
Make DataStore clsoe connections on shutdown, to commit/flush all pending operations

gmt2001:
*** DataStore ***
Added new method CloseConnection, no-op by default

*** MySQLStore, H2Store ***
Add @Override CloseConnection to close the connection to the database

*** SQLiteStore ***
Add @override annotation to CloseConnection

*** PhantomBot ***
Add call to DataStore.CloseConnection just before exiting

*********************

This change will allow the database driver to commit/flush all pending
operations before the bot exits. For example: SQLite is supposed
to commit all changes in the WAL to the db when the last open connection
is properly closed, and then delete the WAL/SHM files (as they are then empty).

*********************

Screenshots before change
https://1drv.ms/i/s!AqJWOcIz2x-UgudbYE_ginXvJ9OKCw
https://1drv.ms/i/s!AqJWOcIz2x-UgudcFhMAbsxNMOrxhg

Screenshots after change
https://1drv.ms/i/s!AqJWOcIz2x-Ugu0Mlm7rEREkaT4GJA
https://1drv.ms/i/s!AqJWOcIz2x-Ugu0DITllDVfZl7RLEQ

Screenshot of the bot launched again and working after the previous screenshot
https://1drv.ms/i/s!AqJWOcIz2x-Ugu0P3Cp-pE8XqaYh0w

*********************

The only thing not pictured in the screenshots is the second commit in this PR which just shuffles around the CRLF at the end of the onExit method to fix the "Closing database" message being on the same line as "Waiting for stuff to finish". I did test the bot again after doing that.

### DIFF
--- a/source/com/gmt2001/datastore/DataStore.java
+++ b/source/com/gmt2001/datastore/DataStore.java
@@ -354,6 +354,9 @@ public class DataStore {
     public Connection CreateConnection(String db, String user, String pass) {
         return null;
     }
+    
+    public void CloseConnection() {
+    }
 
     public void setAutoCommit(boolean mode) {
     }
--- a/source/com/gmt2001/datastore/H2Store.java
+++ b/source/com/gmt2001/datastore/H2Store.java
@@ -106,6 +106,18 @@ public class H2Store extends DataStore {
     }
 
     @Override
+    public void CloseConnection() {
+        try {
+            if (connection != null) {
+                connection.close();
+                connection = null;
+            }
+        } catch (SQLException ex) {
+            com.gmt2001.Console.err.printStackTrace(ex);
+        }
+    }
+
+    @Override
     protected void finalize() throws Throwable {
         super.finalize();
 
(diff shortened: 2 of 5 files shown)


BrandenB:
Looks okay to me, but @IllusionaryOne will have to look at it as well when he gets time to make sure everything is good to him, since we both sign off on pull requests now.

IllusionaryOne:
We are in AutoCommit mode, which, in theory should keep the DB in sync, and the JDBC indicates that at shutdown it does close connections; however, it is up to the remote database software on how to react to the threads and connections.  Obviously, SQLite3 doesn't care, it shuts down when the application shuts down, MySQL might be what cares.  H2 I use in Spring Boot projects and don't have a problem either because we run the server as a thread in the application - in this case, PhantomBot, so it shuts down when the bot shuts down.

Approved, as I can see this being a good thing for MySQL and at least a good thing to do with the other databases.  I should say though, with H2 and SQLite, I hit CTRL-C, CTRL-Z, then kill -9 the bot and have never lost data. :-)

PR #1859, opened 2018-02-02
escape @ when using (followage)

CreativeCodeCat:
**Before:**
Before

**After:**
After


BrandenB:
Looks good, will ask @IllusionaryOne to take a look at it as well.

IllusionaryOne:
Approve.

PR #1864, opened 2018-02-03
Added missing lang entries for discord customcommands

HighHowDoIPlay:
I was unable to reproduce a customapi error (as it just returned null each time) so I cannot submit test cases.

This commit was based off the following bug report: https://community.phantombot.tv/t/missing-lines-in-commands-customcommand-js-for-discord/3308


BrandenB:
Good catch, looks fine on my end, I'll let @IllusionaryOne review it too when he gets a chance.

PR #1868, opened 2018-02-04
Sort Commands & Keywords For Discord

CreativeCodeCat:
Before:
Before Img

After:
After Img


CreativeCodeCat:
This is for the keywordsPanel.js

Before:
Before

After:
After

IllusionaryOne:
Seems good, is similar to how sorting is done elsewhere in JavaScript.

PR #1869, opened 2018-02-05
Fixed errenous output on HTTP DB API getKey

gmt2001:
*** HTTPServerCommon ***
Fixed that the json property "key" was set to "getKey" instead of the actual key that was being queried.

***********************

Before
https://1drv.ms/i/s!AqJWOcIz2x-UgvNRQWsmu4Rh1IlS3Q

After
https://1drv.ms/i/s!AqJWOcIz2x-UgvkfkNrmzmgrM7JF4g


BrandenB:
Looks good, I'll let @IllusionaryOne review it as well and merge if he's okay with it.

IllusionaryOne:
Approved.

PR #1875, opened 2018-02-07
Updated to match pollSystem.js

CreativeCodeCat:
img is here


IllusionaryOne:
Mixed on this, I think "notime" in both spots makes more sense, but "nottime" is ok too.  The idea is that "no time" was provided.  

Leave this for now as 'nottime' in my opinion, can always mull over changing both later.

@ScaniaTV I approve.

PR #1883, opened 2018-02-10
Improved YTP 404 lang, added some spacing for readability.

theTreeSerok:
More informational for both users and admins, in my opinion.


IllusionaryOne:
It isn't clear that song requests are disabled though, how about:

+$.lang.register('ytplayer.client.404', 'The YouTube Player is currently not being used and song requests are disabled.');

We can disable the repeat of a song being requested, so, it might be good to be clear that the song request didn't go through.

@ScaniaTV any additional thoughts?

theTreeSerok:
That works as well.

BrandenB:
I am fine with this change.

PR #1884, opened 2018-02-15
Created TwitchTitleChangeEvent based TwitchGameChangeEvent

sparten9999:


IllusionaryOne:
Looks reasonable.  I see a couple of code style items though.  

init.js:878 > comment starts off a few columns further to the right than it should be.  
TwitchCache.java:354 > comment starts off a few columns further to the right than it should be.
TwitchCache.java:353 and :359 are duplicate assignments.  I believe :359 is not needed anymore.
TwitchCache.java:360 > the brace looks further to the right than it should be.
TwitchCache.java:468 > looks tabbed instead of 4 spaces?

Please provide test screen shots, even if it is just enabling Debug Mode and watching the events fire.  

Double check our Code Style for how we format the code.  

Cheers

sparten9999:
i fixed the codestyle errors. in regards to the TwitchCache.java:353 and :359 duplicates im not comfortable deleting :359.
i copied the code from the section labeled "Get the game being streamed." on line 311.
that section has the same code in the same spots just edited for title instead of game.
the lines are :319 and :322
please advise

sparten9999:
im still gonna do a few more tests and get the screenshots soon

sparten9999:
here is the test 
3ebce39728792aa26b78921059b38381

source/tv/phantombot/cache/TwitchCache.java
```diff
@@ -459,7 +465,8 @@ public String getStreamStatus() {
     public void setStreamStatus(String streamTitle) {
         forcedStreamTitleUpdate = true;
         this.streamTitle = streamTitle;
-    }
+        EventBus.instance().postAsync(new TwitchTitleChangeEvent(streamTitle));  
+	}
```
IllusionaryOne: This brace looks out of place?  Should line up with the setStreamStatus() method?

IllusionaryOne:
Other than the one brace that looks out of place now, I think it is good - @ScaniaTV would also need to review the change.  Thank you for posting the test, it makes us feel better to know folks are unit testing changes and not just changing code :)  I assume you were changing the title a few times to force the firing.  

I am fine with leaving the duplicate assignment, I will need to go in and clean that up later, along with the other section as you pointed out.

BrandenB:
Except that one brace everything looks good to me. Once we move over to Helix we will probably revamp most of this cache anyways.

BrandenB:
Any reason why you named `getTitleTitle` like this? It could simply be `getTitle`.

sparten9999:
i meant it to be getStreamTitle to make it look nicer. must have typod it. its fixed now. here is another pic
3034c08774a563bc608982fe6d975608

BrandenB:
Missing an `r`. After that I'll merge this.

```
public String getSteamTitle() {
 +        return this.streamTitle;
 +    }
```

sparten9999:
fixed the missing r

PR #1905, opened 2018-03-04
Fix backwards iterating over users

BtbN:
This fixes an exception occurring in line 616 due to the iteration starting one behind the actual length.

### DIFF
--- a/javascript-source/core/permissions.js
+++ b/javascript-source/core/permissions.js
@@ -611,7 +611,7 @@
                 // Cast the user as a string, because Rhino.
                 parts[i] = (parts[i] + '');
                 // Remove the user from the users array.
-                for (var t = $.users.length; t >= 0; t--) {
+                for (var t = $.users.length - 1; t >= 0; t--) {
                     if ($.users[t] !== undefined && $.users[t][0] == parts[i]) {
                         $.users.splice(t, 1);
                         break;


IllusionaryOne:
I am considering rejecting this unless you can prove otherwise.  The length is the total length of the array object, and the count starts at 0 with arrays/objects:

```
illusion% java -jar rhino-1.7.7.jar
Rhino 1.7.7 2015 06 17
js> list = [];

js> print (list.length);
0
js> list.push('hello');
1
js> print (list.length);
1
js> print (list[1]);   // this is why we start at length - 1 //
undefined
js> print (list[0]);
hello
js> list.push('goodbye');
2
js> for (var i = list.length; i >= 0; i--) { print (list[i]); }  // using list.length as proposed //
undefined  // undefined value. //
goodbye
hello

```

In an array iteration, you always do the following (not just in JavaScript; this is in C, Java, and so on):
```
for (i = 0; i < list.length; i++) 
```
Note, less than list.length.  This is for the same reason.

javascript-source/core/permissions.js
```diff
@@ -611,7 +611,7 @@
                 // Cast the user as a string, because Rhino.
                 parts[i] = (parts[i] + '');
                 // Remove the user from the users array.
-                for (var t = $.users.length; t >= 0; t--) {
+                for (var t = $.users.length - 1; t >= 0; t--) {
```
IllusionaryOne, line 614: Please see Conversation notes regarding this.

IllusionaryOne (dismissed):
Please see the notes provided.

BtbN:
What? This is iterating _backwards_. An array goes from 0 to length-1 in any sane language(Some BASIC dialects disagree here, but who uses that anyway).
So an array with 3 elements, this starts at 3, then 2, 1, 0, which is obviously one element too many.
JavaScript does not explode on you when you read beyond array length but just gives you an undefined value as result, as your own example demonstrates.
Still, the code in question does not seem to handle that gracefully, even though it looks like it intends to.
The iteration is definitely one too long though, and this change fixes it to not start at one behind array length.

BtbN:
Just realized, from your example: "using list.length as proposed"
I guess you read the PR the wrong way around? I'm proposing the exact opposite, length - 1, which is hopefully obviously right.

IllusionaryOne:
You are correct, I think I am working too much and can't see colors.  You are adding the "- 1" which is correct.  Sorry about that.

To @ScaniaTV for acceptance.  I sign off on this.

BrandenB:
Looks good to me, was about to comment on this being correct. 

Merging, thanks for the fix!

PR #1918, opened 2018-03-16
StinkingBanana_KeywordCount_Fix

wei0831:
## Summary
- Prevent user from calling (keywordcount another_keyword) to increase another keyword's count.
- Having multiple (keywordcount) in the response will work. It only increment once.

## Modification
**scripts\commands\customCommands.js**
- Modified function tags

**scripts\handlers\keywordHandler.js**
- Modified event ircChannelMessage

#### Prevent user from calling (keywordcount another_keyword) to increase another keyword's count.
**Keyword:** ```test1```
**Response:** ```test1 (keywordcount test4) (keywordcount test5)```
**Expected Behavior:** No keywordcounts were increased.
  
test1

**Keyword:** ```test4```
**Response:** ```test4 (keywordcount test5)```
**Keyword:** ```test5```
**Response:** ```test5 (keywordcount test4)```
**Expected Behavior:** No keywordcounts were increased.
  
test4_5

#### Having multiple (keywordcount) in the response will work. It only increment once.
**Keyword:** ```test2```
**Response:** ```test2 (keywordcount) (keywordcount)```
**Expected Behavior:** Keywordcount only increased by one
  
test2

#### Normal Case
**Keyword:** ```test3```
**Response:** ```test3 (keywordcount)```
**Expected Behavior:** Keywordcount increased by one.
  
test3

### DIFF
--- a/javascript-source/commands/customCommands.js
+++ b/javascript-source/commands/customCommands.js
@@ -204,9 +204,9 @@
         }
 
         if (message.match(/\(keywordcount\s(.+)\)/g)) {
-            var input_keyword = message.match(/\(keywordcount\s(.+)\)/)[1],
+            var input_keyword = message.match(/.*\(keywordcount\s(.+)\).*/)[1],
                 keyword_info = JSON.parse($.inidb.get('keywords', input_keyword));
-        
+
             if ('count' in keyword_info) {
                 ++keyword_info["count"];
             } else {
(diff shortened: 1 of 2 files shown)


BrandenB:
Looks good, I'll let @IllusionaryOne look over it and merge if he's okay with it.

PR #1922, opened 2018-03-17
User Only Commands + (points)

CreativeCodeCat:
image

[03-17-2018 @ 20:13:15.560 GMT] dakoda: !test
[03-17-2018 @ 20:13:15.564 GMT] [CHAT]  this is awesome
[03-17-2018 @ 20:14:08.560 GMT] hecodes2much: !test

### DIFF
--- a/javascript-source/commands/customCommands.js
+++ b/javascript-source/commands/customCommands.js
@@ -4,7 +4,7 @@
         reCustomAPIJson = new RegExp(/\(customapijson ([\w\.:\/\$=\?\&\-]+)\s([\w\W]+)\)/), // URL[1], JSONmatch[2..n]
         reCustomAPITextTag = new RegExp(/{([\w\W]+)}/),
         reCommandTag = new RegExp(/\(command\s([\w]+)\)/),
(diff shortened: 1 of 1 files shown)


CreativeCodeCat:
Also added a variable for (points) 

(@sender) you have (points) (pointname)

poinst image

```
[03-18-2018 @ 00:33:24.429 GMT] dakoda: !test
[03-18-2018 @ 00:33:24.445 GMT] [CHAT] @Dakoda,  you have 2543 Assassin_Coins
```

OR 

(@sender) throws some of there (points) (pointname) into a forge turning them into a precious Material

other screen shot

```
[03-18-2018 @ 00:36:35.003 GMT] dakoda: !test
[03-18-2018 @ 00:36:35.005 GMT] [CHAT] @Dakoda,  throws some of there 2554 Assassin_Coins into a forge turning them into a precious Material
```

javascript-source/commands/customCommands.js
```diff
@@ -398,6 +402,15 @@
             }
             message = $.replace(message, message.match(/(\(gameonly=.*\))/)[1], '');
         }
+        
+        if (message.match(/\(useronly=.*\)/g)) {
+            var user = message.match(/\(useronly=(.*)\)/)[1];
```
BrandenB: Since you're matching here, you don't have to match the string again when replacing.

BrandenB (changes requested):
Just fix that one thing and I'll let @IllusionaryOne look over it and see if he's okay with it.

javascript-source/commands/customCommands.js
CreativeCodeCat: so would that be the same for gameonly as i did it the same code style as that one and it has that?? if so ill also change the gameonly one to

BrandenB: Yes. Not sure why it was done like that to begin with.

PR #1928, opened 2018-03-27
- Updated the panel to use Twitch's new chat

BrandenB:


resources/web/panel/js/dashboardPanel.js
```diff
@@ -623,7 +623,7 @@
             // Load the iframe if it isn't there.
             if ($("#chatsidebar").html().indexOf(getChannelName().toLowerCase()) === -1) {
                 $("#chatsidebar").append("<iframe id=\"chat\" frameborder=\"0\" scrolling=\"no\" onload=\"hideLoadingImage()\"" +
-                                             "src=\"https://www.twitch.tv/" + getChannelName().toLowerCase() + "/chat?popout=\">");
+                                             "src=\"https://www.twitch.tv/embed" + getChannelName().toLowerCase() + "/chat\">");
```
EdSmcr, line 626: You are missing a / after the embed
It should be: 
`"src=\"https://www.twitch.tv/embed/" + getChannelName().toLowerCase() + "/chat\">");`
instead of: 
`"src=\"https://www.twitch.tv/embed" + getChannelName().toLowerCase() + "/chat\">");`

EdSmcr (commented):
You are missing a / after the embed

resources/web/panel/js/dashboardPanel.js
Zackery: Thanks for this, we'll get it fixed before release.

PR #1966, opened 2018-04-24
Update main.js

ghost:
- Added two lang strings for easier translation into other languages

### DIFF
--- a/javascript-source/lang/english/main.js
+++ b/javascript-source/lang/english/main.js
@@ -238,6 +238,8 @@ $.lang.register('commandpause.initiated', 'All commands are disabled for $1.');
 $.lang.register('commandpause.isactive', 'All commands are currently disabled.');
 $.lang.register('common.disabled', 'disabled');
 $.lang.register('common.enabled', 'enabled');
+$.lang.register('common.not.allowed', 'not allowed');
+$.lang.register('common.allowed', 'allowed');
 $.lang.register('common.user-error', 'You must specify a user to target with this command');
 $.lang.register('common.user.404', 'The user "$1" has not visited this channel yet.');
 $.lang.register('common.game.change', 'Changed the current game to: $1!');


vesturo:
will it still work with Translations made bevore the change or do i need to Rebuild the Translation?

ghost:
Vesturo, developers approve PLs of each other for a few hours, my first PL been hanging for a week. So at the moment I’m not sure they are interested in the third party contributions. If someone will approve my PLs, I'll let you know what you need to change in the German translation. P.S. I asked the developers to consider my three pull requests as a whole thing, because I still haven't figured out why some commits join the old PL and some don't.

vesturo:
That would be awsome, thank you!

BrandenB:
Looks good to me, will let @IllusionaryOne and @Zackery check and merge.

We always let most developers look at pull requests before merging so nothing gets missed.

Zackery:
Looks good here.

IllusionaryOne:
Normally I do not approve without test results.  Since this is language only, I am going to assume that it was unit tested though and results are available?  I make very few exceptions to this rule.  I also *ignore* pull requests most of the time without them. Please let me know if the pull_request_template is not displaying when you submit a pull request so I can check that.  Part of that message is:

You must provide test results for your change.  If test results are not provided we will do one of two things. We will ask you for test results or we will just reject the change.  Test results convince us that you performed a unit test and verified your change.

We reserve the right to reject a change for any reason at all.  Typically, we will provide a reason but if we do not, that is at our discretion.  There are some reasons for which we will reject changes, other than the ones given above (but they will be repeated):

- Code style does not match.
- Incompatible license/improper use of license of third party software/API.
- No test results.
- Potential performance issues.
- Design changes that go against the core design philosophy.
- Items with potential to spam or consume the outgoing message queue.
- Poorly architected items.
- Item is supported in chat but no update for the Control Panel.
- Any command that would violate the Twitch Terms of Service or the Terms of Service of any provider.

PR #1982, opened 2018-05-08
Added recording the top cheer

Mitsunee:
Added recording the top cheer to database and text file. This is just a simple addition for people who use the text files to display information on stream. Something could be added to the message in the future, I haven't messed around with the language files yet, so idk. The saved values are in phantombot_streamInfo where I also found latestSub, so that should be fine.


Mitsunee:
As I copied the original code from the live version, my commit seems to delete code on here I've never seen

BrandenB:
I like this, however the `topcheer` variable is never initialized when the script first loads, also, just for code format sake, `topcheer` should use camel casing, like `topCheer` for example and there are some code reverts. Finally, please provide test results with this.

Mitsunee:
I guess I can't update the pullrequest to use the other branch? Oh well...

PR #1983, opened 2018-05-08
Updated Botlist (again)

vesturo:
**Addet the Following Bots. this time with a description what they do**

- **electricalskateboard**

> A Twitch bot that does Host Raffles. is in many channels and actually writes nothing, only monitors if stream is live

- **electricallongboard**

> A Twitch bot that does Host Raffles. is in many channels and actually writes nothing, only monitors if stream is live

- **Streamelements**

> the official bot of the Donation & Alert service with the same name. 

- **Stay_Hydrated_Bot**

> A little friendly bot that reminds you to Drink enough

### DIFF
--- a/resources/addons/ignorebots.txt
+++ b/resources/addons/ignorebots.txt
@@ -12,3 +12,6 @@ curseappbot
 revlobot
 muxybot
 faegwent
+electricalskateboard
+electricallongboard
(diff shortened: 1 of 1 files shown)


vesturo:
can be closed too, will do a new one with all changes i made!

PR #1984, opened 2018-05-08
Updated ignorebots.txt with aditional bots

vesturo:
**Addet the Following Bots. this time with a description what they do**
i don't know what tests you want so if you need anything just message me!

_The other pull-requests can be closed, that one here is up-to-date_

- **electricalskateboard**

> A Twitch bot that does Host Raffles. is in many channels and actually writes nothing, only monitors if stream is live

- **fatmanleg**

> A Twitch bot that does Host Raffles. is in many channels and actually writes nothing, only monitors if stream is live

- **electricallongboard**

> A Twitch bot that does Host Raffles. is in many channels and actually writes nothing, only monitors if stream is live

- **Streamelements**

> the official bot of the Donation & Alert service with the same name. 

- **Stay_Hydrated_Bot**

> A little friendly bot that reminds you to Drink enough

### DIFF
--- a/resources/addons/ignorebots.txt
+++ b/resources/addons/ignorebots.txt
@@ -12,3 +12,8 @@ curseappbot
 revlobot
 muxybot
 faegwent
+electricalskateboard
+electricallongboard
+streamelements
+stay_hydrated_bot
+fatmanleg


IllusionaryOne:
Testing this would be a bit interesting.  It would require having those bots in your channel at some point (or adding to the time/points database tables) and then showing them in the top5 commands and then not.  However, since this is just a text file and not code, I think it is ok - you didn't change the code that parses the file. @ScaniaTV ?

Let us say though, for example, that you add a new command or change an existing command.  We would just ask for some screenshots of Twitch chat or your Console just to show that the command works.  Thanks!

vesturo:
well if @ScaniaTV wants it i could set up a test with the electrical bots they seem to be 24/7 in my channel :D 

That is good to know, will remember it if i ever edit some command or so!

BrandenB:
I've seen most of those bots as well, they seem to randomly join channels, well not StreamElements, and the hydrated bot anyone can send it to any channel, which is weird. Anyway, testing this and showing results is not necessary in my opinion, so merging.

PR #2004, opened 2018-06-03
hostHandler.js typo & bitsHandler.js edge case

shiek:


IllusionaryOne:
I think I am ok with this.  What about the edge case if the message has "bits" more than once?  Not sure if that would happen or not.  Should it be a global replace or is that overkill?

@ScaniaTV @Zackery ?

Zackery:
👍🏼

shiek:
@IllusionaryOne `$.replace` already replaces all occurrences of the string.
https://github.com/PhantomBot/PhantomBot/blob/91607a4daec91891d13f3bf76e40d02e3b11b7cc/javascript-source/core/misc.js#L675

BrandenB:
Looks good to me, merging.

PR #2006, opened 2018-06-08
added "Commanderroot"

vesturo:
he's a statistic bot which never watches the stream, but lurks in chat all the time in some channels.

From his panels:
"Why are you in my chat?
No special reason. I lurk in a "few" channels but I don't write anything. I also don't watch the stream."


BrandenB:
I can approve of this, commanderroot is an old staff member and he likes to collect a lot of data. I don't think this requires testing.

https://twitch-tools.rootonline.de/

PR #2042, opened 2018-09-12
Add a Discord audit log for custom commands.

rlazarus:
We currently use the moderation log setting to create an audit log in Discord for Twitch bans and timeouts, but we'd also like to have a record of who creates, edits, and deletes custom Twitch chat commands, along with when. This PR adds a toggle to include those events in the mod log channel (default off).

Tested manually, with the setting both on and off.

**customCommands.js**
- Initialize the `customCommandLogs` setting to `false` if unset
- Add an unexported function `logIfEnabled()` which formats and sends a Discord embed if the setting is turned on
- Call it from each command handler that alters a custom command

**discord.html**
- Add a corresponding toggle to the control panel

### DIFF
--- a/javascript-source/commands/customCommands.js
+++ b/javascript-source/commands/customCommands.js
@@ -1197,9 +1197,9 @@
             if (args.length === 1) {
                 $.say($.whisperPrefix(sender) + $.lang.get('customcommands.reset.success', action));
                 $.logCustomCommand({
-                  'reset.command': '!' + action,
-                  'reset.count': 0,
-                  'sender': sender,
+                    'reset.command': '!' + action,
+                    'reset.count': 0,
+                    'sender': sender,
                 });
                 $.inidb.del('commandCount', action);
             } else {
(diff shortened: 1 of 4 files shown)


BrandenB:
Hey there,

I like the concept of this idea, however I feel like it should be its own module and not just apart of the custom commands module. I wouldn't mind merging this if you made it its own module with a global function that would log in Discord if enabled, or log into a file if it isn't. 

If you do end up doing this, please make sure you avoid as many DB calls as you can, line `#709` could be a variable and not a DB call, also, please be sure to follow our code style. For example, your `if` statement should always use braces `{ }`. Also, every output of the bot should be in a lang file.

Thank you for submitting this.

rlazarus:
Thanks for the feedback! This is ready for another look.

Logging into a file is a good idea -- as long as we have it, I figure we might as well _always_ log the changes there, not just when Discord logging is disabled. What do you think?

BrandenB:
Hey there, I just noticed something, correct me if i'm wrong, but I don't see a way to set the channel to where the logs should be posted to.

BrandenB:
Ah, I see that now. I'll probably add the option to use a different one for the control panel that's being worked on, because I know some people might get confused.

Anyway, thanks again for this.

rlazarus:
Yeah, I decided to go with the simplest possible thing, but a separate setting makes sense too. And thanks again for the review!

PR #2048, opened 2018-10-06
Fix for (pay) tag

PhantomBotRU:
(paycom) tag must be (pay) as below in this js (see lines 286-287)


IllusionaryOne:
The only left over item has nothing to do with code, I don't see this tag (or (paycom) for that matter) in our Command Variables page in Community.  Just have to add it there.

PR #2050, opened 2018-10-12
Feature/discord emotes

Nekroido:
This change allows a bot to be connected to multiple servers while working only with a defined one as usual. Twitch sub emotes are unavailable to bots (with no ETA on implementing the support), so bot owners can bypass this limitation by mirroring their emotes on personal servers. Discord bots have Nitro benefits, so they can use custom emotes from any connected server.

**PhantomBot.java**
- Added `discord_server` config property.

**discord/DiscordAPI.java**
- Added support for connecting to multiple guilds.

### DIFF
--- a/source/tv/phantombot/PhantomBot.java
+++ b/source/tv/phantombot/PhantomBot.java
@@ -418,7 +418,11 @@ public final class PhantomBot implements Listener {
 
         /* Set the Discord variables */
         this.discordToken = this.pbProperties.getProperty("discord_token", "");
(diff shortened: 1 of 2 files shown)


BrandenB:
I don't see a point in merging this in the main build just for emotes. Also, the extra print line you added to the console just adds more of a cluster, it doesn't need to be there. Some of your code doesn't follow ours standards, for example some of your if statements are missing a space after the last `{`.

PR #2061, opened 2018-11-01
Fixed log misprint & quote marks

PhantomBotRU:
Minor improvements:
- fixed misprint for logs;
- replaced double quote marks `"` with single quote marks `'` according to PB code style;
- replaced `can't` with `can\'t`.


BrandenB:
This looks good, we normally require test results but I don't think it's needed in this case.

I'll let @IllusionaryOne look over it and merge if he's okay with it.

PR #2062, opened 2018-11-05
Improved regex for fake purges and CAPS

PhantomBotRU:
### **FAKE PURGES**

At the moment regex `/^<message \w+>|^<\w+ deleted>/i` catches only messages starting with `<message` and/or containing `deleted>`. Examples: `<message deleted>` as it is, `<mEsSaGe annihilated>` and `<My message was unfairly deleted> WTF?`. The last two are definitely not fake purge, but a joke (or attempts to make a joke) or wonder.

Also Twitch added a different purge message for its mobile applications, and it looks like this: `<Message deleted.>`.

My suggestion is not just to add the possible dot before the `>`, but to improve regex this way:
`/(^<[mмμ][eеε]ss[aаα]g[eеε]\s+d[eеε][liι][eеε][tтτ][eеε]d\.?>$)/i`.

Transcription:

- `^<`… – `<`… must be at the beginning of a message;
- `[mмμ]` – any of this letters: Greek uppercase `Μ` (lowercase `μ`) and Cyrillic uppercase `М` (lowercase `м`) looks absolutely similar to Latin `M` (`m`);
- `[eеε]` – any of this letters: Greek uppercase `Ε` (lowercase `ε`) and Cyrillic uppercase `Е` (lowercase `е`) looks absolutely similar to Latin `E` (`e`);
- `ss` – just letters;
- `[aаα]` – any of this letters: Greek uppercase `Α` (lowercase `α`) and Cyrillic uppercase `А` (lowercase `а`) looks absolutely similar to Latin `A` (`a`);
- `g` – just letter;
- `[eеε]` – see above;
- `\s+` – one or more spaces between the first and the last (second) words;
- `d` – just letter;
- `[eеε]` – see above;
- `[liι]` – any of this letters: Greek uppercase `Ι` (lowercase `ι`) and Latin uppercase `I` (lowercase `i`) looks similar to Latin lowercase `l` (uppercase `L`) in many fonts;
- `[eеε]` – see above;
- `[tтτ]` – any of this letters: Greek uppercase `Τ` (lowercase `τ`) and Cyrillic uppercase `Т` (lowercase `т`) looks absolutely similar to Latin `T` (`t`);
- `[eеε]` – see above;
- `d` – just letter;
- `\.?` – null or one dot `.` (for `<message deleted>` in a browser and `<Message deleted.>` in the mobile app);
- …`>$` – …`>` must be at the end of a message.

You may not have seen the substitution of letters from non-Latin alphabets, but it is quite a well-known phenomenon on European channels (Greece, ex-USSR, ex-Yugoslavia, Bulgaria, and even Germany and Israel with huge Russian and Ukrainian communities). Why only Greek and Cyrillic writing systems? Because other [world popular writing systems](https://en.wikipedia.org/wiki/List_of_writing_systems) have only slightly similar, but not absolutely similar letters.
After such improvement the script catches official Twitch messages `<message deleted>` and `<Message deleted.>` in any language combinations of letters and in any case (upper or lower), but does not catches messages like before (examples: "`<message unreasonably deleted>`" or `<message deleted> why???` or even "`<sh*t deleted> FFS!!! Kappa`") that actually are not fake purges but jokes or insults (which are not a duty of fake purge filter but a duty of a moderators).

-------------------------------

### **GREEK AND CYRILLIC CAPS SUPPORT**

`[A-ZА-ЯЁΑ-Ω]` instead of `[A-Z]`

Transcription:

- `Α-Ω` – Greek letters;
- `А-ЯЁ` – Cyrillic letters and Russian letter `Ё` which is not a part of the range `А-Я`.

Why only Greek and Cyrillic again? Other world most popular writing systems doesn't see the difference between uppercase and lowercase.

Thus, `[^a-z0-9 ]` for `nonAlphaCount` and `nonAlphaSeq` should be similarly replaced with `[^a-z0-9а-яёα-ϖ ]` containing lowercase Greek and Cyrillic letters.

-------------------------------

### **TESTING**

I've been testing all this for three weeks: CAPS and fake purge filters work successfully for both monolingual and multilingual messages.
Sorry for my imperfect English.


BrandenB:
To add to what @IllusionaryOne said, right now PhantomBot will remove non English characters, if we remove that feature, users may get upset because PhantomBot was previously moderating non English characters. I do plan on adding a toggle for this once we fully finish the new beta panel. For this reason and the reasons mentioned by @IllusionaryOne I am going to close this pull request.

I will be adding something similar in the future - it's on my todo list.

PR #2084, opened 2018-12-20
Fixes the time System if expired

CreativeCodeCat:
ImageUrl

### DIFF
--- a/javascript-source/core/timeSystem.js
+++ b/javascript-source/core/timeSystem.js
@@ -188,7 +188,7 @@
 
             // If the array is empty, return 0 seconds.
             if (timeStringParts.length === 0) {
-                return ('0 ' + $.lang.get('common.time.seconds'));
+                return ($.lang.get('common.time.expired'));
             }
 
             // Join the array to make a string.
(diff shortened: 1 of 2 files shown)


IllusionaryOne:
No, that is incorrect.  This could be used in a tag for (countdown) after a sentence and would look silly if there was an out of place capital letter.

```
addcom left Days until Something (countdown=Dec 10 2018)
[12-20-2018 @ 20:47:01.421 MST] [MUTED] @IllusionaryBot, Command !left has been created!
left
[12-20-2018 @ 20:47:02.779 MST] [MUTED] Days until Something looks like that countdown has expired
```

vs

```
[12-20-2018 @ 20:47:02.779 MST] [MUTED] Days until Something Looks like that countdown has expired
```

If anything it could be changed to "...and it looks like that countdown has expired" or passed as an optional string to countdown to replace the default lang.  Either way, because it is used conversationally in a tag, the language entry is a placeholder.  Still is better than "0 seconds" that it was displaying.  Folks can easily change the language as needed.

CreativeCodeCat:
I just noticed this is also used for the (countup=Dec 10 2019) as well so maybe it needs to check what one its doing 

fixed the check please [Click Here](https://github.com/PhantomBot/PhantomBot/pull/2085)

PR #2085, opened 2018-12-21
Fixed the counters to display different if it counts up or down

CreativeCodeCat:
this will make the countdown+ display different to the countup if its not met the same dates... example below

```
(countdown=December 20 2018 23:59:59)
(countup=December 20 2019 23:59:59)
```

Img Here

also added the new true and false to the places that need the hoursOnly variable.

Anything else that needs changing just let me know here and ill fix it... Thanks

### DIFF
--- a/javascript-source/commands/customCommands.js
+++ b/javascript-source/commands/customCommands.js
@@ -188,7 +188,7 @@
                 countdown, time;
             countdown = t.replace('(countdown=', '').replace(')', '');
             time = (Date.parse(countdown) - Date.parse($.getLocalTime()));
-            message = $.replace(message, t, $.getTimeString(time / 1000, false));
+            message = $.replace(message, t, $.getCountString(time / 1000, false));
         }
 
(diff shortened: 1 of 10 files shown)


CreativeCodeCat:
Just to add to this... would it be better to set this back to what it use to be and make a new function for the counters as other functions use this like the !top10 and a few others which if no numbers set it should say 0 seconds

Please let me know below on what you think i should do?

CreativeCodeCat:
I believe this is a much better way to do it as it does not break the other 7 scripts that use **$.getTimeString** please let me know what you guys think?

IllusionaryOne:
I agree it makes sense to not break the others.  Something I hadn't considered either.  The language entry is subjective, since it can be inline.  I think with the new panel allowing language updates the string is acceptable.  I'll let this sit for a couple of days to see if anyone else comments else will merge it.

IllusionaryOne:
I lapsed on this one.  Were there any objections? Otherwise, I am good to merge.

PR #2094, opened 2019-01-03
added discord logs / file logs for twitch delete message command

CreativeCodeCat:
*Changes*
 * Added delete to the TwitchPubSub.java
 * Created PubSubModerationDeleteEvent.java class file
 * Added PubSubModerationDelete Hook to init.js
 * Added PubSubModerationDelete Bind to moderation.js
 * Added embedDelete Function to moderation.js

*Logs*
Image Here
Image Here

### DIFF
--- a/javascript-source/discord/core/moderation.js
+++ b/javascript-source/discord/core/moderation.js
@@ -144,6 +144,31 @@
     function timeoutUser(message) {
         $.discordAPI.deleteMessage(message);
     }
+    
+    /*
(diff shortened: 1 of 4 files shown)


IllusionaryOne:
I had to revert a previous change.  If that is fixed and resubmitted before this is merged, then init.js is fine.  If not, then init.js needs to be changed back to what it was before I made changes to support a previous PR.

PR #2105, opened 2019-01-05
Forgot one of the gamewisp settings

CreativeCodeCat:


BrandenB:
Looks good. No need for testing.

PR #2106, opened 2019-01-05
Moved ignore bot commands to permissions

CreativeCodeCat:
*changes*
* Deleted ignoreHandler.js as i have moved them into the permissions.js
* Deleted handlers-ignoreHandler.js as i moved them into main.js

**Note:** The reason for this change is so the the commands work along side the settings for permissions there for if you add a bot to the ignore list it will auto refresh the botlist and not have to run !reloadbots. if you would like i could make a gif of this working and upload if needed


IllusionaryOne:
Should addTwitchBot() clean out the DB entries too?

CreativeCodeCat:
saveBotList() triggers cleanTwitchBots() too so adding or removing will trigger it this way even if someone was to add bots to the file itself then noticed there was a command to remove them and happened to remove it then it would then  still trigger to clear the other new entries :)

IllusionaryOne:
Please test with the Console command `jointest [some_name]` and verify that bots are not added to $.users.  

Thanks!

CreativeCodeCat:
```
jointest coebot
[01-06-2019 @ 23:58:19.768 GMT] [CONSOLE] Executing jointest
users
[01-06-2019 @ 23:58:24.063 GMT] [CHAT] @CruxedBot, Users in channel: cruxedbot, dakoda
```

Note: cruxedbot is not in my list as i have it so you can fight the bot so he needs points

IllusionaryOne:
@ScaniaTV if you approve, I am good.  My primary concern was the change around the data object model for the list of bots, and that is working for jointest.

BrandenB:
The bot `cruxedbot` is in the list, I feel like this system should auto-magically detect real bots (not just check if the name ends with bot).

IllusionaryOne:
I think the only way we could auto-detect is the Known Bot or Verified Bot flags.  However, are all bots registering for that.  Are there other ways to detect bots?

CreativeCodeCat:
we would also need a whitelist for people that want there bot to gain currency or id have to edit my script to allow the bot to fight without points

PR #2109, opened 2019-01-14
New Quote System Feature: Alias Names

Mitsunee:
EDIT: **Do not merge yet**. I'm going to integrate this into the panel and update this PR.

Added support for alias names for quotes from quoteSystem.js

**quoteSystem.js**
- updated function `getQuote()` to attempt to fetch the ID of the quote based on the argument, should it not be a number. Should the alias not exist it will return an empty array, same as if the argument was an invalid ID
- updated function `deleteQuote()` to remove aliases for the quote that is getting deleted and shift down values past the ID of the deleted quote to match the new IDs
- added command `!quoteaddalias [quoteId] [alias]` to add new aliases via chat
- added command `!quotedelalias [alias]` to remove an alias via chat

**Database**
- added file `quotealiases` to store quote aliases. The alias name is used as key, while the ID of the quote it belongs to is stored as the value. This allows for multiple aliases per quote and prevents overlapping names.

**EDIT:** ~~I'll be moving the strings in the added commands to a lang file. Can't test that currently as twitch chat is too slow to work right now.~~ Changes are commited and tested now.

Test results:

quoteaddalias_1  
`!quote` now supports aliases. Adding aliases works and an exception is given should an alias of that name already exist.

quotedelalias_1  
`!quotedelalias` works and gives an exception should an alias of that name not exist.

quotedel_1  
The updated `!delquote` correctly shifts down IDs in the database file for quotes with IDs bigger than the deleted quote. The same result has also been replicated for deleting the command via the beta-panel!

quotealiaserrors_1  
Further exceptions for incorrect or missing arguments for `!quoteaddalias` and `!quotedelalias`

### DIFF
--- a/javascript-source/lang/english/systems/systems-quoteSystem.js
+++ b/javascript-source/lang/english/systems/systems-quoteSystem.js
@@ -34,3 +34,9 @@ $.lang.register('quotesystem.quotemessage.success', 'Changed the message used fo
 $.lang.register('quotesystem.searchquote.usage', 'Usage: !searchquote [text] (Must provide at least 5 characters)');
 $.lang.register('quotesystem.searchquote.404', 'No matching quotes found.');
 $.lang.register('quotesystem.searchquote.found', 'Quote IDs with matches: $1');
+$.lang.register('quotesystem.aliases.add.usage', 'Usage: !quoteaddalias [quoteId] [alias]');
+$.lang.register('quotesystem.aliases.add.duplicate', 'Alias "$1" already exists.');
+$.lang.register('quotesystem.aliases.add.success', 'Added alias "$1" for quote ID $2.');
+$.lang.register('quotesystem.aliases.del.usage', 'Usage: !quotedelalias [alias]');
+$.lang.register('quotesystem.aliases.del.404', 'Could not find alias "$1".');
+$.lang.register('quotesystem.aliases.del.success', 'Deleted alias "$1".');
--- a/javascript-source/systems/quoteSystem.js
+++ b/javascript-source/systems/quoteSystem.js
@@ -78,13 +78,13 @@
             $.inidb.del('quotes', quoteId);
             quoteKeys = $.inidb.GetKeyList('quotes', '');
             aliases = $.inidb.GetKeyList('quotealiases', '');
-            
+
             $.inidb.setAutoCommit(false);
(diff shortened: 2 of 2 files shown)


Mitsunee:
tested another thing:

quotedel_2
Deleting a quote correctly removes associated aliases, so a new quote with the same ID will not inherit old aliases

Mitsunee:
Testing for the 6 new lang strings:
test6

BrandenB:
Thank you for submitting this, however, there's already a way to do this without creating another command system, which can be confusing to some users.

`!aliascom myquote quote 1`

```aliascom myquote "quote 1"
[01-15-2019 @ 08:29:42.876 AST] [CHAT] @SlammerBot, The command !quote 1 was successfully aliased to !myquote
addquote test
[01-15-2019 @ 08:29:53.543 AST] [CHAT] Added new quote from SlammerBot with id #0.
addquote test 2
[01-15-2019 @ 08:30:00.351 AST] [CHAT] Added new quote from SlammerBot with id #1.
myquote
[01-15-2019 @ 08:30:03.921 AST] [CHAT] [1] "test 2", by SlammerBot (15-01-2019)
```

Mitsunee:
> Thank you for submitting this, however, there's already a way to do this without creating another command system, which can be confusing to some users.
> 
> `!aliascom myquote "quote 1"`
> 
> ```
> [01-15-2019 @ 08:29:42.876 AST] [CHAT] @SlammerBot, The command !quote 1 was successfully aliased to !myquote
> addquote test
> [01-15-2019 @ 08:29:53.543 AST] [CHAT] Added new quote from SlammerBot with id #0.
> addquote test 2
> [01-15-2019 @ 08:30:00.351 AST] [CHAT] Added new quote from SlammerBot with id #1.
> myquote
> [01-15-2019 @ 08:30:03.921 AST] [CHAT] [1] "test 2", by SlammerBot (15-01-2019)```
> ```

command aliases do however not automatically get updated, should a quote get deleted and the alias can't start with !quote the same way that using an ID does.

BrandenB:
True, I'll need to review this with the team and try to implement it into the new beta panel.

BrandenB:
After talking about this a bit, we'll be rejecting it since it's more of a specific-channel thing.

PR #2110, opened 2019-01-28
Block Access to commands for ignored bots

CreativeCodeCat:
This will block !gift !points add, take, give and set from adding removing or setting the points of a ignored bot but this does not stop the beta-panel from doing so

**Before:**
image goes here

**After:**
image goes here

### DIFF
--- a/javascript-source/core/permissions.js
+++ b/javascript-source/core/permissions.js
@@ -902,7 +902,7 @@
         }
 
         /**
-         * @commandpath ignoreremove - List the bots from the ignorebots.txt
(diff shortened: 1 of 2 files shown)


CreativeCodeCat:
Should I close this?

IllusionaryOne:
No, I just had 3 weeks where I had conferences for work most of the time so didn't have time to review.  Let me compare to the other PR that was submitted that has a lot of other files in it to make sure we don't cause a conflict down the road (give me a day!)

PR #2112, opened 2019-02-18
Performance improvement for Data Render Service.

rlazarus:
Currently, the Times API is computationally intensive because it makes one DB lookup to get the list of keys in the `time` table, and then another three `time` lookups for _each_ key to get its value. These database calls are the performance bottleneck. (The other parts of the DRS handler do something similar, but the `time` table is largest by orders of magnitude -- at least in my case, and it seems reasonable that that would be common.)

This change adds a `GetKeyValuesList()` method to the `DataStore` interface, so that we can fetch the keys and values in a single DB lookup. On my machine (SQLite DB, 153k rows in the `phantombot_time` table) this reduces the total execution time from 40 minutes to 15 seconds.

Note to reviewers: In order to keep the change small and easily reviewable, I haven't modified the other parts of the DRS handler -- it would speed them up a little, but they aren't as performance-critical since the tables are much smaller. Happy to go back and convert them for consistency, or leave them, whichever way you like.

I also only have a SQLite DB available to test with real-world data, so I've only superficially tested the other three DB types. Do you have test data available?

**javascript-source/handlers/dataServiceHandler.js**
- For the Times data, call `GetKeyValueList()`, and iterate over the array of `KeyValue`s, instead of calling `GetKeyList()`.

**javascript-source/systems/ranksSystem.js**
- Add an optional `time` parameter to `hasRank()` and `getRank()` so that we can skip the DB lookup if we've already done it, but still use the rest of the ranks system logic. The behavior is unchanged if `time` is unset, so other call sites can be left as-is.

**source/com/gmt2001/datastore/Datastore.java**
- Add a `GetKeyValueList()` method, with the same arguments as `GetKeyList()`, to retrieve keys and values simultaneously.

**source/com/gmt2001/datastore/KeyValue.java**
- New immutable data type for String key-value pairs. An array of `KeyValue`s is the return type from `GetKeyValueList()`.

**source/com/gmt2001/datastore/H2Store.java**
**source/com/gmt2001/datastore/IniStore.java**
**source/com/gmt2001/datastore/MySQLStore.java**
**source/com/gmt2001/datastore/SqliteStore.java**
- Implementations of `GetKeyValueList()` for each database type.

### DIFF
--- a/javascript-source/handlers/dataServiceHandler.js
+++ b/javascript-source/handlers/dataServiceHandler.js
@@ -24,6 +24,7 @@
      */
     function drsTimer() {
         var keys,
+            entries,
             parts,
             apiStatus,
             commandValue,
@@ -112,20 +113,21 @@
         apiStatus = $.dataRenderServiceAPI.postData(jsonStringer.toString(), $.channelName, 'points');
         $.log.event('DataRenderService: Points API status : ' + apiStatus);
 
-        keys = $.inidb.GetKeyList('time', '');
+        entries = $.inidb.GetKeyValueList('time', '');
         jsonStringer = new JSONStringer();
         ranksJsonStringer = new JSONStringer();
         jsonStringer.object().key('times').array();
         ranksJsonStringer.object().key('ranks').array();
-        for (var idx in keys) {
+        for (var idx in entries) {
+            var user = entries[idx].key, time = entries[idx].value;
             jsonStringer.object();
-            jsonStringer.key('user').value(keys[idx] + '');
-            jsonStringer.key('time').value($.inidb.get('time', keys[idx]));
+            jsonStringer.key('user').value(user);
+            jsonStringer.key('time').value(time);
             jsonStringer.endObject();
 
             ranksJsonStringer.object();
-            ranksJsonStringer.key('user').value(keys[idx] + '');
-            ranksJsonStringer.key('rank').value($.getRank(keys[idx]));
+            ranksJsonStringer.key('user').value(user);
+            ranksJsonStringer.key('rank').value($.getRank(user, parseInt(time)));
             ranksJsonStringer.endObject();
         }
         jsonStringer.endArray().endObject();
--- a/javascript-source/handlers/dataServiceHandler.js
+++ b/javascript-source/handlers/dataServiceHandler.js
@@ -23,11 +23,9 @@
      * @function drsTimer
      */
     function drsTimer() {
-        var keys,
-            entries,
+        var entries,
             parts,
             apiStatus,
-            commandValue,
             commandHelpData = {},
             commandHelpFileData,
             lastTime = parseInt($.getSetIniDbNumber('datarenderservice', 'last_time', 0)),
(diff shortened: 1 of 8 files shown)


BrandenB:
Hey there, 

This looks really good, we knew this service had a huge performance impact for channels with a lot of data but we never got around to applying a fix for it. We really appreciate the work you've put into this.

If you don't mind adding the use of this new method to the other parts of the DRS it would help us a lot, if not I'll merge anyways and add it later.

Thanks again!

rlazarus:
Done -- thanks for such a quick response!

BrandenB:
Looks good, thanks a bunch!

IllusionaryOne:
@rlazarus Might need to roll this back, it causes errors, please investigate as this was your patch.
```
[03-24-2019 @ 04:38:02.316 GMT] org.mozilla.javascript.EvaluatorException: The choice of Java method java.lang.String.replace matching JavaScript argument types (function,string) is ambiguous; candidate methods are: 
    class java.lang.String replace(char,char)
    class java.lang.String replace(java.lang.CharSequence,java.lang.CharSequence) (dataServiceHandler.js#73)
    at org.mozilla.javascript.DefaultErrorReporter.runtimeError(DefaultErrorReporter.java:77)
    at org.mozilla.javascript.Context.reportRuntimeError(Context.java:998)
    at org.mozilla.javascript.Context.reportRuntimeError(Context.java:1053)
    at org.mozilla.javascript.Context.reportRuntimeError4(Context.java:1040)
    at org.mozilla.javascript.NativeJavaMethod.findFunction(NativeJavaMethod.java:468)
    at org.mozilla.javascript.NativeJavaMethod.findCachedFunction(NativeJavaMethod.java:262)
    at org.mozilla.javascript.NativeJavaMethod.call(NativeJavaMethod.java:139)
    at org.mozilla.javascript.optimizer.OptRuntime.call2(OptRuntime.java:42)
    at org.mozilla.javascript.gen.dataServiceHandler_js_168._c_drsTimer_2(dataServiceHandler.js:73)
    at org.mozilla.javascript.gen.dataServiceHandler_js_168.call(dataServiceHandler.js)
    at org.mozilla.javascript.ContextFactory.doTopCall(ContextFactory.java:405)
    at org.mozilla.javascript.ScriptRuntime.doTopCall(ScriptRuntime.java:3508)
    at org.mozilla.javascript.gen.dataServiceHandler_js_168.call(dataServiceHandler.js)
    at org.mozilla.javascript.Context$1.run(Context.java:533)
    at org.mozilla.javascript.Context.call(Context.java:544)
    at org.mozilla.javascript.Context.call(Context.java:531)
    at org.mozilla.javascript.JavaAdapter.callMethod(JavaAdapter.java:577)
    at adapter1.run(<adapter>)
    at java.util.TimerThread.mainLoop(Unknown Source)
    at java.util.TimerThread.run(Unknown Source)
```
The issue more than likely is the data type pulled back from the new call:
```
        entries = $.inidb.GetKeyValueList('command', '');
        jsonStringer = new JSONStringer();
        jsonStringer.object().key('commands').array();
        for (var idx in entries) {
            var command = entries[idx].key, commandValue = entries[idx].value;
```
When previously:
```
                commandValue = String($.inidb.get('command', keys[idx]));
```
As commandValue.replace() is where the issue is now in the code:
```
                    commandValue = commandValue.replace(reCustomAPI, '(customapi)');
```

IllusionaryOne:
As I said, this is broken.  Will be rolling back to the script change in the next 24 hours.

rlazarus:
Sorry, I've been busy this week -- planning to take a look over the
weekend, but if you need to roll back in the meantime, do what you gotta
do. Sorry for the breakage!

IllusionaryOne:
Things happen, I have rolled it back for now for the Nightly!  Just the script, I have not touched anything else!

Thanks!

PR #2115, opened 2019-02-21
Added getMonths() method for regular/prime sub events

ingostrohmeier:
The `getMonths()` method now also returns the value that is provided by Twitchs `msg-param-cumulative-months` tag upon a regular subscriber event.
Tested with actual subscribers on my channel (using a dev version of SubscriberLevels) and works as expected.


BrandenB:
Looks good, small changes, up to you @IllusionaryOne

PR #2116, opened 2019-03-04
Declare `time` at the beginning of the rank system script

ohcibi:
The error might not happen on some js environments because of the
missing `"use strict";` directive. To avoid such errors in the future it
is advised to put this at the beginning of *all* the other js files as
well.

### DIFF
--- a/javascript-source/systems/ranksSystem.js
+++ b/javascript-source/systems/ranksSystem.js
@@ -25,6 +25,7 @@
 
     var rankEligableTime = $.getSetIniDbNumber('settings', 'rankEligableTime', 50),
         rankEligableCost = $.getSetIniDbNumber('settings', 'rankEligableCost', 200),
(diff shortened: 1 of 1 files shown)


BrandenB:
If a variable isn't declared it will always error no matter what now do to a change on how we manage the global scope in PhantomBot. Using the keyword strict in Rhino JavaScript doesn't do anything like it does in NodeJS.

ohcibi:
@ScaniaTV yes thats why I added the fix about the missing variable declaration. I had a chat with one of your team mates on discord and he told me that he wasn't able to reproduce the problem with current master, so I wrongly concluded that a missing `"use strict";` was causing this confusion.

This very PR however is about fixing said error; so I removed the `"use strict";` and force pushed.

BrandenB:
Looks good now, thanks.

PR #2128, opened 2019-04-01
Roll forward #2112 with the EvaluatorException fixed.

rlazarus:
This fixes the issue described at https://github.com/PhantomBot/PhantomBot/pull/2112#issuecomment-475930097.

`commandValue` used to be a JavaScript String, but #2112 inadvertently made it
a Java String, which broke the call to its `replace()` method -- obviously the
two classes have different APIs. This puts back the cast to JS String.

### DIFF
--- a/javascript-source/handlers/dataServiceHandler.js
+++ b/javascript-source/handlers/dataServiceHandler.js
@@ -23,10 +23,9 @@
      * @function drsTimer
      */
     function drsTimer() {
-        var keys,
+        var entries,
             parts,
             apiStatus,
-            commandValue,
             commandHelpData = {},
             commandHelpFileData,
             lastTime = parseInt($.getSetIniDbNumber('datarenderservice', 'last_time', 0)),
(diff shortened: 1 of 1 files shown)


IllusionaryOne:
Should I test with my PhantomBot that has the tags that blew it up or did you have some (customapi) tags in your custom command list?  I don't mind double checking it - I just want to ask.

Cheers

BrandenB:
Could you also provide console logs to make sure everything works. Thanks! :)

rlazarus:
I don't have any `(customapi)` tags in my regular DB -- that's how this slipped through in the first place. So I added a bogus one to test with, `!addcom !apitest (customapi foobar)`. Good enough to match `reCustomAPI` and make sure the string replacement works.

@IllusionaryOne If you wouldn't mind testing that'd be great -- I'm pretty confident based on my tests but I'd feel even better if you retest with the data we choked on originally.

@ScaniaTV Sure, [here you go](https://gist.github.com/rlazarus/2d23b94bf0a7e16645201b4ed4afa06a) -- for completeness I also included the slightly tweaked dataServiceHandler.js I've been testing with. Since we only need to test the serialization and not the actual API, it comments out all the lines that actually send the assembled JSON to the server and also disables the rate limit, so that I don't have to wait 90 minutes to test.

Thanks for taking a look!

BrandenB:
Looks good to me, thank you for updating it. I'll let @IllusionaryOne look over it once more and merge if he's okay with it!

PR #2130, opened 2019-04-04
Updated launch.bat to use the full path to the jar file

gmt2001:
This patch makes the Windows launch.bat script use the full path to the jar file in the **java** launch line. This will allow users to properly use **Run as Administrator**

Resolves *Error: Unable to access jarfile PhantomBot.jar* when using **Run as Administrator**

Tested with both regular launch and Run as Admin launch

### DIFF
--- a/resources/launch.bat
+++ b/resources/launch.bat
@@ -27,6 +27,7 @@ IF %ERRORLEVEL% NEQ 0 (
     pause
     exit
 )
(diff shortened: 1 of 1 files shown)


BrandenB:
We normally require screenshots to make sure that everything still works like it should, but this is such a small change I don't think that it's need. I am fine with this, I'll let @IllusionaryOne decide if he wants to merge it.

BrandenB:
Will merge anyway since this has been here for a while and it's a small change.

PR #2131, opened 2019-04-06
Updated text from Hosters -> VIPs for group changes

AevumDecessus:
Cleans up the dropdowns in the panel for permissions to reflect
the new VIP group name for group id == 5

command_permission
permission_permissions
audio_hook_permission


BrandenB:
Looks good, simple enough change will merge.

PR #2132, opened 2019-04-13
Fixed Points Remove

CreativeCodeCat:
https://community.phantombot.tv/t/bug-viewers-can-remove-points/6604/8


IllusionaryOne:
Simply adds '!points remove' which was missing.  I approve this, it is a minor change, even without test results.

PR #2137, opened 2019-04-28
Attempted Patch for Promote System Crash on profile_banner

IllusionaryOne:
**promoteSystem.js**
- Stack trace may indicate that the variable 'profile_banner' does not exist when there isn't a banner.
- Checking to see if variable exists now, with a default value of null applied.


IllusionaryOne:
Build failed due to Oracle JDK issue complaining that wanted version 9-13 but found 8 (we use 8).  Committing anyway.

PR #2144, opened 2019-05-06
Fixed To Work Correctly

CreativeCodeCat:
fixed it so it accepts the link the first time by adding 201 and also made it get the Url with the http:// included
img

### DIFF
--- a/source/com/illusionaryone/BitlyAPIv4.java
+++ b/source/com/illusionaryone/BitlyAPIv4.java
@@ -110,6 +110,8 @@ public class BitlyAPIv4 {
 
             if (urlConn.getResponseCode() == 200) {
                 inputStream = urlConn.getInputStream();
+            } else if (urlConn.getResponseCode() == 201) {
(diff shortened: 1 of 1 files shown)


IllusionaryOne:
The "link" is not required:
```
{
  "created_at": "1970-01-01T00:00:00+0000",
  "id": "bit.ly/2WsYrNc",
  "link": "http://bit.ly/2WsYrNc",
  "custom_bitlinks": [],
  "long_url": "https://www.illusionaryone.tv/",
  "archived": false,
  "tags": [],
  "deeplinks": [],
  "references": {
    "group": "https://api-ssl.bitly.com/v4/groups/Bg56fvSTpUC"
  }
}
```
It is the same as ID and worked fine in my testing, it adds length to it though.  The "id" mimics the previous behavior of Google URL Shortener and reduces what is in chat.  Please put that back.

IllusionaryOne:
Talked with HeCodes2Much, says that Discord requires the 'http://' so merging anyway!

PR #2147, opened 2019-05-10
Added Manual and Automatic Commercials

gmt2001:
I went back into my old PhantomBot code to lookup how I did automatic commercials and rewrote it into the current bot as its own system.

I additionally updated the TwitchAPIv5.RunCommercial method to properly execute the way the API documentation expects (send JSON instead of form encoded) and updated the JavaDoc info for it

This system is operated like so:

**A.** The command _!commercial (length) [silent]_ can be used (default moderators) to instantly run a commercial of the specified length. If the _silent_ option is not present, the bot will announce in chat _Running a $1 second commercial_

**B.** The command _!commercial autotimer (interval) (length) [message]_ can be used (default admin) to setup the autotimer to run commercials automatically. If the _message_ parameter is left out, the bot will not announce anything in chat.

The autotimer can be disabled by using _!commercial autotimer off_ and some commands are included to change/delete the message (technically you could also just issue the full _!commercial autotimer_ command at any point to change parameters)

When the autotimer is set it will run the first automatic commerical _length_ minutes after the time it was programmed, or _length_ minutes after bot startup (whichever is later)

Unfortunately I do not have a partnered channel or work with any partnered streamers that I could use to test out the commercial system, so if you guys can find someone willing to help us test it, I would appreciate that. I am confident that there are no issues on my end of the code, unless there is a minor typo somewhere (I did triple check for them).

### DIFF
--- a/javascript-source/systems/commercialSystem.js
+++ b/javascript-source/systems/commercialSystem.js
@@ -27,16 +27,16 @@
      * @function startCommercialTimer
      */
     function startCommercialTimer() {
+        lastCommercial = $.systemTime();
+        
         interval = setInterval(function() {
-            lastCommercial = $.systemTime();
-            
             if (commercialTimer && $.bot.isModuleEnabled('./systems/commercialSystem.js')) {
                 if ((lastCommercial + (commercialInterval * 6e4)) <= $.systemTime()) {
                     if ($.isOnline($.channelName)) {
-                        $.twitch.RunCommercial($.channelName, commercialLength);
+                        var result = $.twitch.RunCommercial($.channelName, commercialLength);
                         lastCommercial = $.systemTime();
                         
-                        if (commercialMessage.length() > 0) {
+                        if (commercialMessage.length() > 0 && result.getInt("_http") !== 422) {
                             $.say(commercialMessage);
                         }
                     }
--- a/javascript-source/systems/commercialSystem.js
+++ b/javascript-source/systems/commercialSystem.js
@@ -174,5 +174,38 @@
             startCommercialTimer();
         }
     });
+
+    /**
+     * @event webPanelSocketUpdate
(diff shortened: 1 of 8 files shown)


javascript-source/systems/commercialSystem.js
```diff
@@ -0,0 +1,173 @@
...
+                    if (args.length < 2 || !args[1].equalsIgnoreCase("silent")) {
+                        $.say($.whisperPrefix(sender) + $.lang.get('commercialsystem.run', args[0]));
+                    }
+                }
+            }
+        }
+    });
+
+    // Set the interval to run commercials
+    if (commercialTimer) {
```
BrandenB: Would you mind moving this to the `initReady` event, then it would only start the timer if the module is enabled, or gets enabled.

BrandenB (changes requested):
Hey gmt2001, 

I'm glad that you've decided to add this module to the bot, a lot of partnered streamers have been asking for it mostly due to Twitch mandating a certain amount of commercials ran per stream.

Just to keep everything user-friendly, would you mind adding a tab under the extra section of the control panel to update these settings, maybe having an option on the main dashboard as well to run an instant commercial (be sure to follow the current codding standards in the panel). 

Thanks!

BrandenB (approved):

BrandenB:
Looks good, only thing the module toggle doesn't seem to be working for me, maybe it's an issue on my end? Everything else seems fine though, just maybe align the buttons for the update/disable commercial to the right if you don't mind.

gmt2001:
Ready for final review

BrandenB:
Looks all good to me, tested it and everything seems to work.

Going to let @IllusionaryOne do another rewrite to make sure everything looks good and merge it.

PR #2148, opened 2019-05-10
Added HTTP Basic Auth support to HttpRequest

gmt2001:
This patch adds HTTP Basic Auth support to com.gmt2001.HttpRequest, and by extension to (customapi) and (customapijson)

Link to the demo's I used to test this: https://jigsaw.w3.org/HTTP/

TestResult

### DIFF
--- a/source/com/gmt2001/HttpRequest.java
+++ b/source/com/gmt2001/HttpRequest.java
@@ -56,6 +56,11 @@ public class HttpRequest {
 
             HttpURLConnection h = (HttpURLConnection) u.openConnection();
 
+            if (u.getUserInfo() != null) {
+                String basicAuth = "Basic " + new String(new Base64().encode(u.getUserInfo().getBytes()));
+                h.setRequestProperty("Authorization", basicAuth);
+            }
+
             headers.entrySet().stream().forEach((e) -> {
                 h.addRequestProperty(e.getKey(), e.getValue());
             });
--- a/source/com/gmt2001/HttpRequest.java
+++ b/source/com/gmt2001/HttpRequest.java
@@ -22,6 +22,7 @@ import java.net.URL;
 import java.util.HashMap;
 import java.io.BufferedInputStream;
 import java.io.BufferedOutputStream;
+import org.apache.commons.codec.binary.Base64;
 import org.apache.commons.io.IOUtils;
 
 /**


BrandenB:
Looks good, small enough change I am okay with it.

I'll let @IllusionaryOne merge if he's okay with it as well.

IllusionaryOne:
I have an overall problem with it.  We should add a feature to encrypt the username:password.  I can imagine a moderator looking at the command and obtaining credentials.  I understand that API keys can be visible, but API keys don't potentially provide account access.

Either we should have a way for the user to encrypt the data somehow or allow Base64 encoding (which is standard for things like this on an Authentication header).  For example could tell the user go to https://www.base64encode.org/ and put in the string for authentication.  Then, make sure that the Command List feature with the Data Render Service is still hiding this information and it is somehow also blocked out when the command contents are viewed.

gmt2001:
I'm just implementing support for a valid uri scheme of http(s)://user:pass@domain/path

If you want I can expand the customCommand.js script to have a _(token)_ subtag for _(customapi)_ that can then be set with _!token command txt_ and replace the subtag at command runtime. And additionally I can attempt to parse out this type of basic auth and replace it with the new subtag on command edit.

IllusionaryOne:
What are your thoughts on it?  I just don't like the idea of some one actually running !addcom from the command line.  I know most people use the Control Panel but some may not.  I am thinking we OFFER a base64encoded option (and if they want to use username:password fine, that is on them).

Maybe something like https://b64(base64-encoded-string-here)@domain ?  Tell them how to get the base64 string? Or a better way?  I don't think a colon can appear in a base64encoded string (maybe it can, I need to check) and if it can't, then just a simple if not contains : then type statement?

Thoughts?

IllusionaryOne:
And yes, I completely agree that is good to offer the option period.  I am just thinking of ways to also protect people.  Sorry to be a pain :-)  Something I could implement actually, just fishing for thoughts on how to best do it.

gmt2001:
I think the token idea I put above is better because then we are not falling into the Security through Obscurity trap. Especially if we restrict read access to the tokens table to require writing a custom script or connecting to the DB yourself (aka panel and chat should be write only for !token unless someone goes out of their way to write a script for it)

IllusionaryOne:
!token sounds good.  I suppose if someone does enter a !token in chat, no one would know what it was for, unless they turn around and immediately add it in a command.  Can you elaborate more on your design idea?

Also, thanks for putting up with my thoughts on this.  I spend a lot of time on security items at work, so it sometimes bleeds over (annoyingly) elsewhere :-)

IllusionaryOne:
All that said, I am merging this for now.  No sense in holding up the train for improvements to be made to the next model.

gmt2001:
Quite simply the new usage for a basic auth in a custom command (you could also use this for a token based really) would be _!addcom mycmd (customapi http://(token)@example.com)_

Then after that you run _!token mycmd user:pass_ and we store it in a separate _cmdtokens_ table

When a viewer then uses _!mycmd_ and we detect it is a customapi, we will just do a _cmdstring.replaceAll("(token)", $.inidb.get("cmdtokens", cmdname))_ before passing it on to $.customapi to execute the request. After $.customapi runs we blank/null out the cmdstring to erase from memory

IllusionaryOne:
I still wish we weren't typing user:pass into chat.  I suppose people can always add from the Console and run the command as the bot.  I can look at adding to the Control Panel.

PR #2149, opened 2019-05-14
Added (token) subtag for (customapi)

gmt2001:
This patch adds the (token) subtag for (customapi)/(customapijson) as discussed in PR #2148 

I am not 100% that my method of fixing the column width for the panel is acceptable for you guys. I am also not sure if you want it added separately this way or just want it added to the edit dialog. Please let me know.

This tag is only parsed inside of a (customapi) or (customapijson) tag. It's purpose is to hide an HTTP Basic Auth user:pass or an API key from viewers/moderators who may be editing commands from chat (or using the panel, if someone shares it among trusted users against the teams advice).

To add a token from chat (not recommended) or from the bot console, the command is _!tokencom command token_

To add a token from the panel, click the new blue lock icon under the **Actions** column on the **Custom Commands** page

At runtime, if a (token) tag is detected inside of a (customapi) or (customapijson) tag, it will be replaced with the stored token value from the database in the temporary URL string that gets passed into $.customApi

Untitled

The first test (!test3) is using the url _https://guest:guest@jigsaw.w3.org/HTTP/Basic/_ without attempting to use the new (token) subtag

The second test (!test5) is using the url _https://(token)@jigsaw.w3.org/HTTP/Basic/_ but failing to set the token value with !tokencom or the panel

The final test (!test4) is the same as the second test, but this command has had it's token set using the panel

### DIFF
--- a/javascript-source/commands/customCommands.js
+++ b/javascript-source/commands/customCommands.js
@@ -953,7 +953,7 @@
          * @commandpath tokencom [command] [token] - Stores a user/pass or API key to be replaced into a (customapi) tag. WARNING: This should be done from the bot console or web panel, if you run this from chat, anyone watching chat can copy your info!
          */
         if (command.equalsIgnoreCase('tokencom')) {
-            if (action === undefined || subAction === undefined) {
+            if (action === undefined) {
                 $.say($.whisperPrefix(sender) + $.lang.get('customcommands.token.usage'));
                 return;
             }
@@ -981,7 +981,11 @@
                 $.say($.whisperPrefix(sender) + $.lang.get('customcommands.token.success', action));
             }
             
-            $.inidb.SetString('commandtoken', '', action, argsString);
+            if (argsString.length() === 0) {
+                $.inidb.RemoveKey('commandtoken', '', action);
+            } else {
+                $.inidb.SetString('commandtoken', '', action, argsString);
+            }
             return;
         }
 
--- a/javascript-source/commands/customCommands.js
+++ b/javascript-source/commands/customCommands.js
@@ -614,6 +614,10 @@
 
         // Get the URL for a customapi, if applicable, and process $1 - $9.  See below about that.
         if ((regExCheck = message.match(reCustomAPI))) {
+            if (regExCheck[1].indexOf('(token)') !== -1 && $.inidb.HasKey('commandtoken', '', command)) {
+                regExCheck[1] = regExCheck[1].replace(/\(token\)/gi, $.inidb.GetString('commandtoken', '', command));
+            }
+            
             if (regExCheck[1].indexOf('$1') != -1) {
                 for (var i = 1; i <= 9; i++) {
                     if (regExCheck[1].indexOf('$' + i) != -1) {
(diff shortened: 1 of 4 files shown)


BrandenB:
Hey gmt2001,

I am fine with having the token command, however the panel side doesn't go well with me. I feel like most people might think it's a button to enable/disable a command, and we'll get more questions about it than anything else, also I doubt this will be used enough to be on the same page as custom commands. It also looks out of place.

If a user wants to add a token without anyone seeing it, they could whisper the bot, sure anyone who has access to the console logs could see it, or if Twitch can see whispers (they probably can), so this might not work with @IllusionaryOne. 

The one way that I could see doing it on the panel is that if the `(token)` tag is present in the response text area, a new field under advanced would be added for the token, with maybe a link to a forum post on about what it does and how it works in the label of the textbox.

gmt2001:
I put the panel option in because IO specifically said he'd prefer it there in the other PR. The issue is just figuring out where to put it and how to display it

BrandenB:
My thought would be the last thing I mentioned in my last reply, other than that I have no other idea.

Maybe our community manager @Collussus , who's also a web developer might have an idea on how to implement this.

https://imgur.com/a/qAi1d5C
https://imgur.com/a/efBBKeI

itsDave1:
> The one way that I could see doing it on the panel is that if the (token) tag is present in the response text area, a new field under advanced would be added for the token, with maybe a link to a forum post on about what it does and how it works in the label of the textbox.

What Scania described here would be the most user friendly to do it. That lock icon is gonna confuse everyone, especially if it's on every single command. It's better to have it in the commands settings itself.

IllusionaryOne:
> I put the panel option in because IO specifically said he'd prefer it there in the other PR. The issue is just figuring out where to put it and how to display it

Yes I definitely would like a Panel option and Console. I am just concerned of the people out there that might see someone put in a username:password credential in Twitch chat and try to use it.  I know it seems like a lot of hoops, I just want to ensure the safety of our users as much as possible.

gmt2001:
Okay, I came up with something that merges your guys' ideas with mine.

When you edit a command, if the response text matches regex _/\(customapi/gi_, then it adds a link under Advanced labeled _Add/Edit Command Token_, which when clicked will open up my modal dialog on top. Note that this is not available during the initial adding of a command since it would fail the commandExists check.

My reason for using the second modal is because I still think it wiser for this particular item to have the basic documentation be right in front of the user every time they use it instead of having to launch into an FAQ on the forums. For usability purposes, the user should not have to launch a manual every time they want to use a feature where they don't quite remember how it works, when we are on a highly user friendly web control panel.

I already know you guys won't be happy with a modal-on-modal approach, but it's the best way I can think of to keep it user friendly without turning the first modal into a mess.

BrandenB:
I am fine with the modal on modal, I understand your point of view.

I do not like the link part though, I added a button myself with your current code, however I deleted my changes by accident, once I do it again I'll submit it to your branch. It basically adds a button under the response box instead of a link under advanced.

BrandenB:
IO was fine with this, I am happy with the changes, merging.

PR #2135, opened 2019-04-20
Added support for defining a section when performing dbquery requests…

gmt2001:
… through the HTTP API

I defined a variable here to replace the hard-coded section parameter strings that defaults to an empty string, which will provide the same behavior/response as before, however, the user can define a *section* GET parameter to override it and retrieve any section they want from the requested table.

I had to change keyExists to use the Full DataStore API method *HasKey* as part of this implementation.

I will post the tests in a reply here once I have them available for you.

PR #2153, opened 2019-05-17
Fixed blank section bug, while retaining ability to explicitly use it

gmt2001:
Removed some redundant catches in sanitizeOffset
Removed some redundant catches in sanitizeLimit
Suppress FinalizeDeclaration warning (this warning is just a "Hey, you used this, be careful")
Suppress some UseSpecificCatch that we are okay with
Use diamond inference

The main change in this patch is to fix a bug with methods such as GetKeysByOrderInternal that causes a query for a blank section name (the default section used in tables) to instead just return all rows, including those that have made use of the section column, as noted in PR #2135 

I also included some minor suggested standard code changes that were noted by NetBeans and eliminated a few redundant catches in 2 methods that will not cause an issue.

### DIFF
--- a/source/com/gmt2001/datastore/DataStore.java
+++ b/source/com/gmt2001/datastore/DataStore.java
@@ -178,7 +178,6 @@ public class DataStore {
         throw new UnsupportedOperationException();
     }
 
-    @SuppressWarnings("UseSpecificCatch")
     public long GetLong(String fName, String section, String key) {
         String sval = GetString(fName, section, key);
 
@@ -195,7 +194,6 @@ public class DataStore {
         SetString(fName, section, key, sval);
     }
 
-    @SuppressWarnings("UseSpecificCatch")
     public int GetInteger(String fName, String section, String key) {
         String sval = GetString(fName, section, key);
 
@@ -212,7 +210,6 @@ public class DataStore {
         SetString(fName, section, key, sval);
     }
 
-    @SuppressWarnings("UseSpecificCatch")
     public float GetFloat(String fName, String section, String key) {
         String sval = GetString(fName, section, key);
 
@@ -229,7 +226,6 @@ public class DataStore {
         SetString(fName, section, key, sval);
     }
 
-    @SuppressWarnings("UseSpecificCatch")
     public double GetDouble(String fName, String section, String key) {
         String sval = GetString(fName, section, key);
 
--- a/source/com/gmt2001/datastore/DataStore.java
+++ b/source/com/gmt2001/datastore/DataStore.java
@@ -178,6 +178,7 @@ public class DataStore {
         throw new UnsupportedOperationException();
     }
 
+    @SuppressWarnings("UseSpecificCatch")
     public long GetLong(String fName, String section, String key) {
         String sval = GetString(fName, section, key);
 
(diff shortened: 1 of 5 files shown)


gmt2001:
I reused my test case from PR #2135 that had revealed the issue. For reference, the setup is:

> For this test I used the phantombot_updates table since it is already populated with a bunch of data. I hand-modified rows 1-3 to have a section of Alpha, 4 & 5 to have Bravo and 6 to have Charlie for the purposes of this test.

The bug as shown previously caused this result

> Result of http://localhost:25000/dbquery?table=updates&keyExists=installedv2&webauth=token
> 
> >     {"table":{"table_name":"updates","key":"installedv2","keyExists":true}}
> 
> ^ This is an existing bug related to the query we use for HasKey that I just noticed, caused by the if (section.Length() > 0) block in https://github.com/PhantomBot/PhantomBot/blob/master/source/com/gmt2001/datastore/SqliteStore.java#L785

Which was wrong because I had changed the _section_ on that row to _Alpha_, but it should have excluded this row (and therefore returned _keyExists = false_)

Under this change, explicitly defining a blank section in the URL will return the expected result

> Result of http://localhost:25000/dbquery?table=updates&section=&keyExists=installedv2&webauth=token
> 
> >     {"table":{"table_name":"updates","key":"installedv2","keyExists":false}}

Scripts looking for the new behavior will need to use the advanced methods and explicitly define _section_ as an empty string and scripts already using the advanced methods will default to the new behavior. The HTTP DBQuery API will exhibit the old behavior by default as a backwards-compatibility unless the section is explicitly defined as blank as in the above test.

IllusionaryOne:
I disagree with removing specific catches, it is not best practice at any of my employers. I appreciate trying to compact the code. I also appreciate having the exact reason for the failure with the proper exception. As Oracle's guide to Java Exception Patterns illustrates, there is a reason that exceptions are written into classes and when individuals develop classes the exceptions should be meaningful and with purpose.   A good description of this:

---
Don't catch generic Exceptions
Often in complex software, a given block of code executes methods that throw a variety of exceptions. Dynamically loading a class and instantiating an object can throw several different exceptions, including ClassNotFoundException, InstantiationException, IllegalAccessException, and ClassCastException.

Instead of adding the four different catch blocks to the try block, *a busy programmer may simply wrap the method calls in a try/catch block that catches generic Exceptions* (see Listing 3 below). While this seems harmless, some unintended side effects could result. For example, if className() is null, Class.forName() will throw a NullPointerException, which will be caught in the method.

---

Obviously these exceptions are not the same as the ones mentioned above.  It does follow the same principle though.

gmt2001:
In the 2 spots where I was adjusting the Exceptions we were literally just catching 2 specific exceptions followed by the generic and returning the same default value on each one as a catch-all. We did not make any use of the actual exceptions whatsoever.

If we were actually making use of the Exceptions and their information I would certainly have left them as they were. 

See https://github.com/PhantomBot/PhantomBot/blob/master/source/com/gmt2001/datastore/MySQLStore.java#L67 for the referenced code

I actually get an interesting result when I _git blame_ the generic Exceptions in DataStore.java
https://github.com/PhantomBot/PhantomBot/blame/1f238b126d66e4f01b7cc3926774295c1d8e3f24/source/com/gmt2001/datastore/DataStore.java#L187

gmt2001:
Did a partial revert. PR now only contains the bugfix itself

IllusionaryOne:
Thank you for taking the review into consideration.  Understandably, at times, there will be times where there is a conversation around items, I go through the same thing at work as a manager and as a senior developer leading projects.

I appreciate all of the help and feedback.

Cheers

BrandenB:
IO seems okay with this, I am too, if not, a revert can be done later.

PR #2154, opened 2019-05-24
Ticketraffle multiplier fix

thebeardbe:
A very small fix for the ticketraffle.

### Problem:

When a ticketraffle with a x times multiplier for subs or regulars is activated, the sub/regular gets attributed the extra tickets but also get deducted the points for the extra tickets.

**Expected behaviour:**

normal user buys 1 ticket at 1 point and gets charged 1 point and gets 1 ticket
regular/sub buys 1 ticket with 2x multiplier and gets charged 1 point and gets 2 tickets.

**Current behaviour**

normal user buys 1 ticket at 1 point and gets charged 1 point and gets 1 ticket
regular/sub buys 1 ticket with 2x multiplier and gets charged 2 points and gets 2 tickets.

### Fix:

added a extra variable for the cost to calculate the real cost.

Test see screenshot. 

BeardBEBot is running the fix
BeardBEBottest is the latest normal nightly build

phantombot

### DIFF
--- a/javascript-source/systems/ticketraffleSystem.js
+++ b/javascript-source/systems/ticketraffleSystem.js
@@ -178,7 +178,7 @@
         if (!$.inidb.exists('entered', user.toLowerCase())) {
             totalEntries++;
         }
-
+	price = times;
         if (tags.getTags().containsKey('subscriber') && tags.getTags().get('subscriber').equals('1')) {
             times *= subTMulti;
         } else if ($.isReg(user)) {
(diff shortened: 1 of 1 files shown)


BrandenB:
Small fix, I saw a post about this on the forums. I am fine with merging, will let @IllusionaryOne review and merge if he thinks everything is alright as well.

Thanks for this!

thebeardbe:
I have checked the topic in question. I've have not solved that problem. I could look into solving that too, but that might be a bit more complicated than this fix.

This fix is only to fix the double spending of the points when a multiplier is activated. The total amount of previously purchased tickets is not changed.

BrandenB:
I am fine with the single change for now, up to @IllusionaryOne if he's okay with this and merge it.

gmt2001 (approved):

PR #2155, opened 2019-05-25
ANSI Quote Table Names

gmt2001:
This patch ANSI Quotes table names to enable support for dashes in table names.

### DIFF
--- a/source/com/gmt2001/datastore/H2Store.java
+++ b/source/com/gmt2001/datastore/H2Store.java
@@ -151,12 +151,12 @@ public class H2Store extends DataStore {
         // Creates a database with 3 columns, the section and variable are used as keys.  value is a 2GB CLOB of text.
         try (Statement statement = connection.createStatement()) {
             statement.setQueryTimeout(10);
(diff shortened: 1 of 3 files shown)


gmt2001:
Appears to be working in SQLite so far

IllusionaryOne:
I am still inclined to just block dashes in YouTube and put in an exception throw if a table is ever requested with a dash.  There are different ways to handle this in different database engines, and it doesn't seem like a wholly good idea to use dashes.  What also worries me is if another engine is added that has a different way of handling it, and it isn't used at all or doesn't accept this method, that creates an issue out of the gate.

I can be convinced still that this is acceptable, just giving my two cents on why I originally opened a ticket on the private Trello to just block hyphens as a solution and was already developing it for a merge, which would supersede this.

gmt2001:
I do not condone it, but the name validator does so I was just fixing the bug that I pretty much created. The other solution you can use here is to just change `validateFname` to remove dashes as a valid character, since it already replaces invalid characters with an underscore, which is valid.

PR #2160, opened 2019-06-07
New (help=) Custom Command Tag

IllusionaryOne:
**customCommand.js**
- The (help=) tag will produce a help message when no parameters are entered.
- Only use this tag with commands that require a parameter.
- The tag does not check to see if parameters are required, it is assumed that the user configuring the custom command is putting it there if there are parameters to be passed.


IllusionaryOne:
This is helpful for things like this:

```
addcom botcolor .color (1) (help=See the color command for the list of valid colors from Twitch: https://help.twitch.tv/s/article/chat-commands)
[06-07-2019 @ 13:09:46.965 MDT] [MUTED] @IllusionaryBot, Command !botcolor has been created!
botcolor
[06-07-2019 @ 13:09:49.733 MDT] [MUTED] See the color command for the list of valid colors from Twitch: https://help.twitch.tv/s/article/chat-commands
botcolor DodgerBlue
[06-07-2019 @ 13:10:11.267 MDT] [MUTED] .color DodgerBlue
```

PR #2164, opened 2019-06-17
Fixed infinite modal causing divs to block page

gmt2001:
Fixes the bug described in [this](https://community.phantom.bot/t/multiple-dialogue-boxes-can-be-opened-forcing-users-to-refresh-before-being-able-to-continue-using-the-panel/7055) forum post


BrandenB:
Simple fix, doesn't really need a long review.

Will merge @IllusionaryOne, if anything comes up we'll revert

PR #2166, opened 2019-06-28
Added detection of arm devices running OpenJDK (assumed rPi)

gmt2001:
Launcher exits fatally if an arm device runs OpenJDK
Fatal exit provides instructions to switch to Oracle JDK
Added missing fatal exit for missing java condition

I do not own a Raspberry Pi. If one of you has one, please help verify that this detects it properly

### DIFF
--- a/resources/launch-service.sh
+++ b/resources/launch-service.sh
@@ -26,7 +26,18 @@
 
 unset DISPLAY
 
-if [[ $(uname) -eq "Darwin" ]]; then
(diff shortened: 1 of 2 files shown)


gmt2001:
Seems to be working

Note the file not found error on the end is because the tester was only testing the script, without having PhantomBot present

https://cdn.discordapp.com/attachments/247160900115628032/595158851158802442/unknown.png

PR #2168, opened 2019-07-19
Upgrade to Java_WebSocket_1.4.0

gmt2001:
Fixes infinite reconnect loop
Fixes thread sync failure that could lead to multiple connections being open
Fixes .mods output not being properly parsed
TwitchWSIRCParser is now a singleton
Reconnect is now protected from running multiple instances by Lock and boolean


gmt2001:
Also did a bit of code cleanup in the areas I was working on in this PR

BrandenB:
Just to add details to this pull-request, this bug was found by a community I am apart of, after a few hours of testing gmt2001 and I found out that it was possible for the bot to open multiple reconnect threads, which could lead them to all connect at the same time (if a network outage happened in the middle) and open multiple connections to Twitch and send multiple of the same messages, I've tested this fix and it does work.

Merging so people can get the fix.

source/tv/phantombot/twitch/irc/TwitchWSIRCParser.java
```diff
@@ -316,6 +331,7 @@ private void onChannelJoined(String message, String username, Map<String, String
 
         // Join the channel.
         webSocket.send("JOIN #" + channelName);
+        session.send(".mods");
```
BrandenB: .. rm this

BrandenB (changes requested):

PR #2170, opened 2019-07-24
Added exponential backoff to reconnect code

gmt2001:
Changed doReconnect to use a ReentrantLock instead of a boolean

This is a followup to PR #2168 to implement exponential backoff in the reconnect logic as [recommended](https://dev.twitch.tv/docs/irc/guide/#re-connecting-to-twitch-irc) by Twitch. This will prevent massive console spam of reconnects when certain conditions (such as a loss of internet connectivity) are present.

The backoff value starts at 1 second after the first failed reconnect and then doubles each time thereafter, maxing out at 5 minutes. Once a connection has been maintained for 10 minutes (MAX_BACKOFF * 2), the backoff resets.


BrandenB:
Tested and it works fine.

PR #2172, opened 2019-08-08
Updated customapi and customapi JSON tags to be fully compatible with RFC 3986 URIs

gmt2001:
See http://www.faqs.org/rfcs/rfc3986.html sections 2.1, 2.2, and 2.3 for more information about the characters supported by RFC 3986 URI's


gmt2001:
Before


After


JSON file used for test
https://next.json-generator.com/api/json/get/VJRm_qVXD

PR #2187, opened 2019-09-07
Fixed missing NewLine-at-EOF issues in handlers

DelTomix:
Of the 15 source files in the javascript/handlers these 8 had no newline at the end of the file which git will insist on fixing and will clutter future diffs with the otherwise-unrelated correction. The other 7 files did not have this issue so seems correct to fix as such. 

As the diffs show the changes are identical to all of the following files;

javascript-source/handlers/clipHandler.js
javascript-source/handlers/donationHandler.js
javascript-source/handlers/emotesHandler.js
javascript-source/handlers/followHandler.js
javascript-source/handlers/gameScanHandler.js
javascript-source/handlers/keywordHandler.js
javascript-source/handlers/streamElementsHandler.js
javascript-source/handlers/tipeeeStreamHandler.js
 - Fix missing New-line character at end of file.


gmt2001 (approved):

DelTomix:
### Update: Added Commit "Fixed Missing Newline-at-EOF entire source tree"

I hope this is helpful and that I am approaching it correctly. 

While waiting for merge - after noticing other instances I did a scan of all relevant file-types across the entire source tree **I found 208 more files with same issue.** Given this is the case it now seems silly fixing only one sub-directory,  and make more sense to just fix the entire tree. (I was originally focused on the handlers directory only because I happened to be working in there.) Thus I have added a commit to this PR with the exact same change for the other 208 files exhibiting the same issue.  

The scan and fix script I made simply excluded:
- all existing binary file extensions, 
- all files of zero bytes length,
- the .git/* tree

Due to the large number of files changed the list is added below as an attachment. Also I did some basic testing info just to be extra sure.  

Attached here for reference:

- [scanlist-fixed-NL-at-EOF.txt](https://github.com/PhantomBot/PhantomBot/files/3593278/scanlist-fixed-NL-at-EOF.txt) - a full list of files that were affected generated by the fix script. Total 208 files changed (plus the 8 above) out of 772 eligible. 

- [debug-level-buildLogs.diff.txt](https://github.com/PhantomBot/PhantomBot/files/3593303/debug-level-buildLogs.diff.txt) - To be extra careful I built both a stock version from master, and the proposed branch, at debug level (i.e.  ant -d -Dbuildtype=nightly_build ). Let me know if you need full build log. 

- [console-testrun.log](https://github.com/PhantomBot/PhantomBot/files/3593317/console-testrun.log) - console output of the bot built from the proposed PR branch

_For future:_ Since this will likely continue to occur (as various contributors inevitably will use various editors) maybe would be cool if this could be **a validation check**? (for appropriate file types of course) and perhaps a style guideline?

Happy to squash/rename/revise/resubmit as instructed.

gmt2001 (approved):

BrandenB:
Merge conflict, not merging until it is fixed.

Also, apart from what GitHub says, I don't think this matters too much, I can agree to merging it.
Will let @IllusionaryOne take the final decision on it.

IllusionaryOne:
I haven't had Git insist on fixing the files, and while I see no harm in this, I am not sure the need for the change, including uglified/compressed JS files that came that way from the vendor.  That aside, I suppose it is fine.

DelTomix:
> I am not sure the need for the change including uglified/compressed JS files that came that way from the vendor

I agree it makes sense that minified/uglified files should be left as-is since they are typically never edited manually, I will exclude them as well as the vendors directory and update the PR accordingly.
It might be wise/convenient for me to rebase my branch - is that ok? 

So I'm going to elaborate on my original rationale - at least to prove I had one and I'm not a style-nazi or anything. :)  I'm cool either way if it is accepted or not. I still think its a good idea. 

> Also, apart from what GitHub says, I don't think this matters too much

True - It doesn't - much. I understand it seems kind-of nit-picky. Its basically a "best-practice" suggestion, and a consistency issue. There are **no** conditions that having the terminating NL present in editable source files can cause problems or inconveniences, but there are _some_ conditions where _if missing_ it can (at least) be a significant inconvenience - depending on your tools and working environment. (see below)  

> I haven't had Git insist on fixing the files,

Various editors will default to _either_ leave the last character of the file as-is - or - default to automatically add the NL character at the end of the file. In the case that your editor leaves as-is, there is no byte difference for git/diff to address, so all-is-well and as intended. In the case that an editor automatically inserts the newline, **and** where it was not present in the original source file, diff will (correctly!) detect it as a change and add the hunk in the patch file accordingly. Having this added to the patch is not functionally problematic, but it adds clutter and deviates (albeit trivially) from the "one-topic" per patch best practice, and has no relevance to the intended changes. Even if the hunk is edited from the commit, git will keep detecting the DIFFerence and insist on staging/comitting. The only escape is to commit the change or go back and delete the character in the changed file.

In my case I work in a Linux/bash environment, and most Linux editors will automatically add the NL. In the case of my main editor (Vim) it can be disabled, and in others (gedit) it cannot. It is not desirable to disable automatic NL@EOF - as a default - since can affect behavior of various other standard tools such as  wc, grep, sed, shell scripts, prompts, and so on.  Some compilers will also generate warnings. Its STILL not a big deal, just a PITA if I forget to tell my editor to leave as-is when I'm in the project.  

Other reasons;
- Consistency - most of the codebase already has NL@EOF in it anyway, 
- Its a POSIX standard (which may not matter much but many of our tools including git itself come from that world, regardless of what platform you use) 
- There are no downsides I can see, but some upsides. 
- It is widely considered best practice.

IllusionaryOne:
As most people would just say "because I said so" and you took the time to really explain the reasoning behind this - I am in full support.  Approved for merge :-)

PR #2189, opened 2019-09-09
Fixed rogue character that got inserted in hostHistory

gmt2001:


BrandenB:
We usually ask for testing screenshots, but in this case it is not needed, will be merging, don't think I need someone else to look at it.

- If you think otherwise @IllusionaryOne you can revert.

PR #2192, opened 2019-10-04
Fixed pollSystem display in Panel

gmt2001:
Removed useless, buggy string replace

pollSystem was doing a useless, buggy replacement of spaces in poll options with `%space_option%`, which was not properly handled on either side (Twitch, Panel) and caused the pie chart on the panel to appear incorrectly (Strikethrough on options, Votes appearing to be attributed to the wrong option)

This patch fixes it.

NOTE: The OBS Poll Chart was not affected because it followed a PubSub model over the WebSocket that was not using this string replace

Before

After


BrandenB:
Seems good, thank you.

PR #2195, opened 2019-11-05
Updaed header copyright year to include 2019

AevumDecessus:
Just a simple cleanup to include 2019 in the copyright header in all files

Command run: 
```find . \( -type d -name .git -prune \) -o -type f -print0 | xargs -0 sed -i 's/(C) 2016-2018/(C) 2016-2019/g'```


BrandenB:
Thanks!

PR #2200, opened 2020-01-26
new custom command tag: repeat

tixona:
The repeat tag will repeat the given text (second arg) x times (first arg).

first arg: only postive interger
second arg: only a-z and spaces

Examples:
!no1 - (repeat 10,FailFish BabyRage)
!no2 - Noooo (repeat 5,FailFish)

Its a bit like the repeat command from SE

### DIFF
--- a/javascript-source/commands/customCommands.js
+++ b/javascript-source/commands/customCommands.js
@@ -442,10 +442,21 @@
         }
 
         if (message.match(/\(repeat\s[1-9]\d*,[\w\s]+\)/g)) {
+            var MIN_COUNTER_VALUE = 1;
+            var MAX_COUNTER_VALUE = 30;
+
             var matches = message.match(/\(repeat\s([1-9]\d*),([\w\s]+)\)/);
             var allMatch = matches[0];
-            var counter = matches[1];
+            var counter = parseInt(matches[1]);
             var iteratingText = matches[2];
+
+            if (counter < MIN_COUNTER_VALUE) {
+                counter = MIN_COUNTER_VALUE;
+            }
+
+            if (counter > MAX_COUNTER_VALUE) {
+                counter = MAX_COUNTER_VALUE;
+            }
 
             var tmpArray = [];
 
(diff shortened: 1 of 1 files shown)


tixona:
Rebased this pull request so its based on the new master branch

javascript-source/commands/customCommands.js
```diff
@@ -441,6 +441,21 @@
...
 
+        if (message.match(/\(repeat\s[1-9]\d*,[\w\s]+\)/g)) {
+            var matches = message.match(/\(repeat\s([1-9]\d*),([\w\s]+)\)/);
+            var allMatch = matches[0];
+            var counter = matches[1];
+            var iteratingText = matches[2];
+
+            var tmpArray = [];
+
+            for(var i = 0; i < counter; i++) {
```
BrandenB, line 463: The variable `counter` is never parsed as an `number`, meaning this could cause issues for people.

javascript-source/commands/customCommands.js
```diff
@@ -441,6 +441,21 @@
             }
         }
 
+        if (message.match(/\(repeat\s[1-9]\d*,[\w\s]+\)/g)) {
+            var matches = message.match(/\(repeat\s([1-9]\d*),([\w\s]+)\)/);
+            var allMatch = matches[0];
+            var counter = matches[1];
```
BrandenB: Make sure to include a limit here, I would say 10 at max so people don't run into Twitch's chat limit.

BrandenB (changes requested):
Everything looks fine apart from the two things I mentioned.

javascript-source/commands/customCommands.js
tixona: If necessary, I can do it. But the group ([1-9]\d*) actually ensure that there are only positive integers allowed

javascript-source/commands/customCommands.js
tixona: I would not set a limit if a user wants to enter a high value so why not. I currently use 30 myself at its works fine with emots for me. So if its necessary whats about 30?

BrandenB: https://github.com/PhantomBot/PhantomBot/blob/master/development-resources/CODESTYLE.md#chat-limits

javascript-source/commands/customCommands.js
BrandenB: You're still comparing a number to a string doing `i < counter`, it may work fine on your system but can cause issues on other systems.

BrandenB (approved):

PR #2206, opened 2020-03-07
Added blacklisted words are blocked in commands

NgLoader:
Blacklist words are now blocked when you try to execute commands like:
- !hallo (when hallo is blacklisted)

I tested it by my self and it works fine

### DIFF
--- a/javascript-source/core/chatModerator.js
+++ b/javascript-source/core/chatModerator.js
@@ -2015,6 +2015,7 @@
     });
 
     /** Export functions to API */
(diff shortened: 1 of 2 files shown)


BrandenB:
Hello,

We do not accept changes that will slow down how fast commands can be processed. 

With this change, if someone has over 100 blacklisted words we need to walk the entire list before the command can be sent. 

Please be sure to read the contribution guide for any other contributions. 

Thanks,
Branden

PR #2211, opened 2020-03-20
Don't let bot answer it's own questions.

robsdedude:
Imagine you have a timer set up like `command:!poll` to let chat know about the ongoing poll.
This change will avoid the bot to to prefix the command with it's own name like "@botname, there's an ongoing poll…"


BrandenB:
I like this idea, thank you for implementing it.

PR #2219, opened 2020-04-15
Add missing Twitch clip command vars

jaredcat:
Heres proof of work on Twitch: 
image

and on Discord:
image

### DIFF
--- a/javascript-source/discord/handlers/clipHandler.js
+++ b/javascript-source/discord/handlers/clipHandler.js
@@ -42,6 +42,7 @@
 	$.bind('twitchClip', function(event) {
         var creator = event.getCreator(),
             url = event.getClipURL(),
+            title = event.getClipTitle(),
             s = message;
 
         /* Even though the Core won't even query the API if this is false, we still check here. */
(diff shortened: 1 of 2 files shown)


dungi:
Uh nice. Currently the (title) is not working in Twitch Chat. Will this fix this also?

jaredcat:
@dungi yes, check the first screenshot

dungi:
Ah the screenshot was so small, didn't saw that. Nice. Then I can create a new feature after that. Cause I needed the title to check, if the user created the clip accidentally or not. Cause this happen 2-3 times a stream. If the title is the same as the stream title, I can prevent printing the clips to discord and the chat and manually post them to discord, if the content is valid.

Thanks for your work.

jaredcat:
@dungi you should make that a toggleable feature. Users are often just too lazy or just don't know to create a title.  
We curate every single clip on our channel and there are a lot of legitimate clips with the stream title (yes a lot of them are bad or accidental but not all)

dungi:
Yes that was the plan.

gmt2001 (approved):

PR #2222, opened 2020-04-24
Patch 70

gmt2001:
- Fix WsSharedRWTokenAuthenticationHandler not authenticating readauth tokens
- Fix equalsIgnoreCase on undefined and Java vs JS for permissions.getKeyIndex
- Add handling for attempts to connect to HTTP/WS instead of HTTPS/WSS when SSL is enabled
- Add better response when attempting to send an authenticated request to an unauthenticated endpoint
- Fix $.permCom


BrandenB:
Merging it, but not because I trust these changes.

gmt2001:
LUL

PR #2223, opened 2020-04-27
Gameonly timers

jaredcat:
Adds in `(gameonly=)` variable to timers.  
If a notice with `(gameonly=)` is selected it will call `sendNotice()` again for the next notice the interval isn't wasted.  

image

### DIFF
--- a/javascript-source/systems/noticeSystem.js
+++ b/javascript-source/systems/noticeSystem.js
@@ -73,14 +73,23 @@
             CommandEvent = Packages.tv.phantombot.event.command.CommandEvent,
             notice = $.inidb.get('notices', 'message_' + RandomNotice);
 
-        if (notice == null) {
-            return;
-        }
-
         RandomNotice++;
 
         if (RandomNotice >= numberOfNotices) {
             RandomNotice = 0;
+        }
+
+        if (notice == null) {
+            return;
+        }
+
+        if (notice.match(/\(gameonly=.*\)/g)) {
+            var game = notice.match(/\(gameonly=(.*)\)/)[1];
+
+            if (!$.getGame($.channelName).equalsIgnoreCase(game)) {
+                return sendNotice();
+            }
+            notice = $.replace(notice, notice.match(/(\(gameonly=.*\))/)[1], "");
         }
 
         if (notice.startsWith('command:')) {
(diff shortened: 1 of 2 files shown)


javascript-source/systems/noticeSystem.js
```diff
@@ -73,14 +73,23 @@
...
         }
 
-        RandomNotice++;
+        if (notice.match(/\(gameonly=.*\)/g)) {
+            var game = notice.match(/\(gameonly=(.*)\)/)[1];
 
-        if (RandomNotice >= numberOfNotices) {
-            RandomNotice = 0;
+            if (!$.getGame($.channelName).equalsIgnoreCase(game)) {
+                return sendNotice();
```
gmt2001: This seems subject to an infinite loop if all notices are `gameonly` tagged and you are not playing any of the tagged games

gmt2001 (changes requested):

javascript-source/systems/noticeSystem.js
jaredcat: I thought about that. I assumed people would have at least one hit.  
What do you suggest? Don't call `sendNotice()` again?

gmt2001: Maybe use a loop around the function instead of recursive calling
This way you can have a var to track if you've tried them all

javascript-source/systems/noticeSystem.js
```diff
@@ -71,16 +71,30 @@
...
+            
+            if (notice && notice.match(/\(gameonly=.*\)/g)) {
+                var game = notice.match(/\(gameonly=(.*)\)/)[1];
+                if ($.getGame($.channelName).equalsIgnoreCase(game)) {
+                    notice = $.replace(notice, notice.match(/(\(gameonly=.*\))/)[1], "");
+                } else {
+                    notice === null;
+                }
+            }
+        } while(!notice || start !== RandomNotice);
```
gmt2001: Still a potential infinite loop, change to `&&`

gmt2001 (changes requested):

gmt2001 (approved):

BrandenB (approved):

PR #2224, opened 2020-04-27
Add toggle to disable checking names for quotes in the database

jaredcat:
Adds dropdown in the quote settings to disable strict checking of usernames. 
When disabled allows for any name to be used.  
Defaults to true 
image
image


gmt2001 (changes requested):
Recommend adding a chat command for this toggle as well

jaredcat:
> Recommend adding a chat command for this toggle as well

Let me know if you have suggestions on naming or output string text

gmt2001 (approved):

BrandenB (approved):

PR #2246, opened 2020-05-20
Add case matching option to keyword system

### LINKED ISSUE #2327: [Enhancement] Ignore case and spaces checkbox for Keywords
Just a suggestion to implement ignore case for keywords which makes it easier for people like me who wants the bot to respond to phrases without having to write a sentence in 20891838 versions all over again just that the bot reliably responds to all keywords.

For example as regex keywords I use 

(What are your specs|What PC do you have|What's your CPU)

but these do not trigger if said person does not use the exact versions as above, one types pc in lowercase, the other one in uppercase... or when the spaces are off. That's why I had to manually go into the keywordHandler script in position 35 and edit

json.regexKey = new RegExp(json.keyword.replace('regex:', ''));
to
json.regexKey = new RegExp(json.keyword.replace('regex:', ''), 'i');

for the cased letters.

(Thanks to a friend)

robsdedude:
This PR solves two minor bugs and adds one feature (all closely related, that's why it's one PR):
 - the temporal variable `str` in `keywordHandler.js` was never cleared. So `str.includes(…` would get slower, the more keywords there were.
 - `resources/web/panel/js/pages/keywords/keywords.js` would throw an error if you changed the keyword (i.e. key) of an existing keyword and tried to edit it again without reloading
 - It gives the user the option to choose, both for regex and regular keywords, if the should be match case sensitive or insensitive.

Problems/things to discuss:
 - This is not backwards compatible. It will break existing regex keywords because:
   - I removed the `regex:` prefix from the key of regex keywords as it was redundant and I couldn't come up with a way to extend that design idea to incorporate the case sensitivity flag.
   - Regex keywords were case sensitive by default. To make it more consistent I choose to make them insensitive by default.
 - I'm not sure if I like the `--regex` and `--case-sensitive` flags the way they are. Using bash like arguments feels out of line with the rest of the bot's modules. But again, I couldn't come up with a better way of implementing this functionality in the chat interface.

Note, that all problems/discussion points only stem from the third point of things this PR solves. So even if you decide to reject the PR, please consider working in the other two minor fixes.

I've tested the matching and the chat interface in my twitch chat and it worked flawlessly. The panel controls also worked as they are supposed to.

### DIFF
--- a/javascript-source/core/updates.js
+++ b/javascript-source/core/updates.js
@@ -877,9 +877,9 @@
         $.inidb.set('updates', 'installedv3.0.1', 'true');
     }
 
-    /* version 3.2 updates */
-    if (!$.inidb.exists('updates', 'installedv3.2') || $.inidb.get('updates', 'installedv3.2') != 'true') {
-        $.consoleLn('Starting PhantomBot update 3.2 updates...');
+    /* version 3.3.0 updates */
+    if (!$.inidb.exists('updates', 'installedv3.3.0') || $.inidb.get('updates', 'installedv3.3.0') != 'true') {
+        $.consoleLn('Starting PhantomBot update 3.3.0 updates...');
 
         $.consoleLn('Updating keywords...');
         var keys = $.inidb.GetKeyList('keywords', ''),
@@ -917,8 +917,8 @@
             $.inidb.set('keywords', newKeywords[i].key, JSON.stringify(newKeywords[i].json));
         }
 
-        $.consoleLn('PhantomBot update 3.2 completed!');
-        $.inidb.set('updates', 'installedv3.2', 'true');
+        $.consoleLn('PhantomBot update 3.3.0 completed!');
+        $.inidb.set('updates', 'installedv3.3.0', 'true');
     }
 
     /**
--- a/javascript-source/core/updates.js
+++ b/javascript-source/core/updates.js
@@ -877,9 +877,9 @@
         $.inidb.set('updates', 'installedv3.0.1', 'true');
     }
 
-    /* version 3.1.3 updates */
-    if (!$.inidb.exists('updates', 'installedv3.1.3') || $.inidb.get('updates', 'installedv3.1.3') != 'true') {
-        $.consoleLn('Starting PhantomBot update 3.1.3 updates...');
+    /* version 3.2 updates */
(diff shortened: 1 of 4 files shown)


robsdedude:
I just learnt about `javascript-source/core/updates.js`. So I will add a migration for the regex keywords. But before I do this, I'd like to get some feedback of you on how you like the idea as well as the current implementation.

robsdedude:
Since this will probably go under in discord, I'll cross post:

> So I'm currently working on adding a case sesitivity switch to keywords. While doing so, I removed the very redundant regex: prefix of regex keywords. I'm now writing the db migration and there is a slight change (pretty unlikely) that this migration causes problem for users: If a users has the exact same keyword as regex and as regular keyword (e.g. test and regex:test) I have to decide what to do. Here are some ideas:
>  - Make the migration fail and tell the user to resolve the conflict. To do so they'd either need to use a sqlite editor or downgrade the bot to start it and use the chat or web panel interface.
>  - Just rename one of the occurences (e.g. appending a number) and hope, that the user notices that the keyword does not work anymore. Even when I ouput a warning to stdout, users that have the bot running as service will not notrice right away (they would with solution 1)
>  - Any other ideas are welcome.
> 
> So what would you say is the best way to handle this?

Currently option 1 is implemented and tested in this PR. If you are fine with it and rest of the design decisions, you can merge. If not, please give me feedback.

gmt2001 (approved):

BrandenB (changes requested):
Next version will be 3.2.0

javascript-source/core/updates.js
```diff
@@ -877,9 +877,9 @@
         $.inidb.set('updates', 'installedv3.0.1', 'true');
     }
 
-    /* version 3.1.3 updates */
-    if (!$.inidb.exists('updates', 'installedv3.1.3') || $.inidb.get('updates', 'installedv3.1.3') != 'true') {
-        $.consoleLn('Starting PhantomBot update 3.1.3 updates...');
+    /* version 3.2 updates */
```
BrandenB: 3.2.0

https://semver.org/

BrandenB (changes requested):

javascript-source/core/updates.js
robsdedude: Yeah, that slipped through. I know what semantic versioning is… Guess I've missed 3.2.0, so what will be the next version I should aim for for this feature?

gmt2001 (approved):

PR #2249, opened 2020-05-25
$.discord.addRole -> $.discord.setRole

robsdedude:
As written here: https://community.phantom.bot/t/role-management-does-not-work/8710
`$.discord.addRole` does not exists and causes an error.


robsdedude:
So this has been fixed by https://github.com/PhantomBot/PhantomBot/pull/2250/files#diff-a1d356cdb4e567bbf43a9ffb9577ec0eL160-R160 as well. However I'd argue, that this PR is cleaner as it channels all the setRoll/addRoll calls through the javascript function which DRYs up the code and makes it easier to adjust, in case discordApi's interface changes in the future.

PR #2251, opened 2020-05-26
WIP: Robsdedude fix custom command random variable

robsdedude:
Currently, if you have multiple random variables in you custom commands (e.g. `(#)` or `(# 1,2)`) phantom bot only draws one random number per same random variable string. 

Example: Custom command `(# 1,6) | (# 1,6) | (# 1,7)` will result in `a | a | b` where `a` is one random number from 1 to 6 and `b` another random number from 1 to 7.

### DIFF
--- a/javascript-source/commands/customCommands.js
+++ b/javascript-source/commands/customCommands.js
@@ -128,7 +128,9 @@
                 var path = RegExp.$2;
                 var path2 = RegExp.$1;
                 var results = $.arrayShuffle($.readFile('./addons/' + path.trim().replace(/\.\./g, '')));
-                message = $.replace(message, '(' + path2.trim(), $.randElement(results));
+                message = $.replaceFunc(message, '(' + path2.trim(), function() {
+                    return $.randElement(results);
+                });
             }
         }
 
@@ -271,7 +273,9 @@
 
         if (message.match(/\(random\)/g)) {
             try {
-                message = $.replace(message, '(random)', $.username.resolve($.randElement($.users)));
+                message = $.replaceFunc(message, '(random)', function () {
+                    return $.username.resolve($.randElement($.users));
+                });
             } catch (ex) {
                 message = $.replace(message, '(random)', $.username.resolve($.botName));
             }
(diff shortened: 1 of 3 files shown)


robsdedude:
I'll update the PR to also fix https://community.phantom.bot/t/multiple-readfilerand-in-one-command/8745 and to optimize run time. Testing for that is pending.

javascript-source/commands/customCommands.js
```diff
@@ -301,13 +307,19 @@
...
-        if (message.match(/\(# (\d+),(\d+)\)/g)) {
-            var mat = message.match(/\(# (\d+),(\d+)\)/);
-            message = $.replace(message, mat[0], String($.randRange(parseInt(mat[1]), parseInt(mat[2]))));
+        if (message.match(/\(# (\d+),(\d+)\)/)) {
+            var mat = message.match(/\(# (\d+),(\d+)\)/),
+                min = parseInt(mat[1]),
+                max = parseInt(mat[2]);
+            message = $.replaceFunc(message, mat[0], function () {
+                return String($.randRange(min, max));
+            });
```
taradinoc, line 322: I think this will fail for messages that contain multiple `#` tags with different parameters, like `You open chest number (# 1,10), revealing (# 100,500) gold coins`: only the `(# 1,10)` will be replaced.

(Same goes for the `readfilerand` case above, I think. It'll handle multiple reads of the same file, but not different files.)

robsdedude: Indeed. I had another version where that kind of issue would not have happened, but I did some optimization and broke it by doing so :/
Thanks for checking and letting me know.

robsdedude: Looking at the code base it seems that all custom command variables with argument(s) have that exact problem. @gmt2001, is this designed like this on purpose? If not, I'd extend this PR to also fix this.

As example `(code=2)-(code=3)` would be expanded to e.g. `7j-(code=3)`.

gmt2001: Was for performance in the past

robsdedude: Does "In the past" mean it can be changed or should it stay as is?

gmt2001: If you can make it better, go for it

gmt2001:
Resolved in PhantomBot/PhantomBot#2253

PR #2252, opened 2020-05-30
WIP: Fix code variable would only work by itself

robsdedude:
Currently the code variable e.g. `(code=3)` would only be properly substituted if nothing else is in the response.


robsdedude:
Marked as WIP, waiting for #2253.

gmt2001:
Resolved in PhantomBot/PhantomBot#2253

PR #2253, opened 2020-06-01
Rewrite custom command variable expansion

robsdedude:
The past days, I've been working on rewriting the variable expansion in custom commands. After trying to patch issues with #2251 and #2252, I came to the conclusion that a proper rethinking would be appropriate. If you merge this, you can close both aforementioned PRs, as this one fixes both problems.  
This change might be something for the next major version update, as it breaks backwards compatibility of some variables. I'll list details further down. First, let me pitch you 

### The gist
Custom commands get checked for anything that looks like a variable to expand, e.g. `(var args)`, `(var=args)`, `(var|args)`, `(var)`, `(var )`, and more would match. After a match is found, the corresponding transformer in `transformers` is looked up and called. Transformers get the variable's arguments (everything after `var`) and the event as arguments and, as their name suggests, process and transform (i.e. expand) the variable. Transformers return objects with one or more of the following properties:
 - `result`: a string with which to replace the variable
 - `cache`: if the result can be cached for the variable with that exact arguments. cache only spans the lifetime of one command call.
 - `raw`: if the `result` should not be escaped (more on escaping further down)
 - `cancel`: if `true`, processing gets immediately stopped and no (further) output is generated
A transformer can also return a falsey value, if the variable should be left untouched.
Variables are expanded from inside to outside and leftmost to right most. This means nested variables work flawlessly.

### Escaping
If a variable's args should contain `(`, `)` or `\` the must be escaped by prepending `\` if they otherwise could be confused with variables. Example: `(readfile nice\(1\).txt)` to read from file `nice(1).txt` while `(readfile nice(1).txt)` would read from file `nice(1).txt` where `(1)` gets replaced with the command's first argument.

### Run time analysis
I didn't perform actual timing, but theoretically (given that object property lookup is efficiently implemented, e.g. with hashmaps) this approach should be faster most custom commands. Before, the string was checked against all possible variables. Now, it's checked against what looks like any variable and then a lookup in an object + a function call happens.

### Detailed changes
Old behaviors that have changed and might make a major release necessary:  
 - customapi / customapijson:
    - `$n` was not expanded, unless `$1`-`$(n-1)` existed
    - if `$1` was expanded to `$2` it would be expanded again
    - if url returned `(command ...)` it would actually get expanded. Note: you can still get this behavior by wrapping the api variable with a `(unescape …)`  
 - commands with `(followage)` do not pass first two arguments as (user, channel) to variable by default. use `(followage (1) (2))` for the same behavior.  
 - commands with `(hours)` do not pass first argument as user to variable by default. use `(hours (1))` for the same behavior.  
 - commands with `(pointtouser)` and `(touser)` do not pass first argument as user to variable by default. use `(pointtouser (1))` or `(touser (1))` for the same behavior.  
 - `(points)` now optionally takes a user argument  
 - `(pay)` and `(price)` now optionally take a command as argument  
 - `(follows)` doesn't add a space after the number anymore
 - `(# min:int,max:int)` was changed to `(# min:int, max:int)` for a more consistent design  
same with `(repeat n:int,message:str)` -> `(repeat n:int, message:str)`
 - `(titleinfo)` and `(gameinfo)` considers all white-space-only strings as not set. Before only '' and ' ' were considered as no title/game set.
 - `(readfile ...)`, `(readfilerandom ...)`, and `(writefile ...)` will now let you know when the file was not found.
 - default args like `(1=random)` now also work up to 9.
 - `(3)` now also gets expanded, even if `(1)` and/or `(2)` do not exist. Similar with other numbers.
 - gracefully handles `(keyword k:str)` when `k` does not exist.
 - `a: (commandslist)` and `(commandslist)a: ` produced the same output :`a: <here went the command list>`. `a: (commandslist)` will now produce the same output as `(commandslist)` and as `(commandslist)` produced before. Use `(commandslist a: )` to get the same result as `a: (commandslist)` produced before.

### Tests
I've tested all variables, but only while I was offline. I'm also not in a twitch team. Hence, I could only test the `(team_…)` variables very rudimentarily.

If you have any questions about this, pleas don't hesitate to ask. I think this could be big step towards a cleaner code base and custom commands system.


BrandenB:
I like this rewrite, however I have one question. Will this still be compatible with keywords? I know from the past that using "command only" variable could cause an error when used in keywords, maybe implementing a way to differentiate the two would be a good add-on as well.

robsdedude:
I didn't know that the keyword system was tied to the custom command system. I'll look into this and do some tests in the next days.

robsdedude:
@BrandenB I've looked into the code and couldn't find anything that might cause a problem. But the code base is huge so I might have just missed it. Could you point me into a direction e.g. with a test case or something that could potentially be problematic?

gmt2001 (dismissed):

gmt2001 (approved):

gmt2001:
Copyedited the OP, as it will be linked to as the expanded release notes for this merge

PR #2262, opened 2020-06-08
Updated help text for alert and playsound tags

AevumDecessus:
Updates the help text in panel/alerts to reflect the newly added (alert) and (playsound) tag support added in b675b843012d

Also corrects the help text for single gifted subscriptions, and a few grammatical corrections for the other messages.

Tests:
image
image
image
image
image


BrandenB:
We might need to be more specific with how the tags works, maybe include it in the tooltip.

example (not sure how the tag works so don't take this as a good example):
`(name), (reward), (playsound soundname.mp3)`

AevumDecessus:
I thought about that, but for example the (alert) tag has [1 required and 4 optional parameters](https://github.com/PhantomBot/PhantomBot/blob/master/resources/web/alerts/js/index.js#L241-L261), so wasn't sure about the best way to include that much information

PR #2287, opened 2020-06-22
Adding search function to quote system

robsdedude:
This allows you to write `!quote keyword` and the bot would return any random quote that contains the given keyword (search is case insensitive). If the keyword is not found, a random quote will be returned.

Tested and couldn't find a bug.


robsdedude:
In discord, the point came up that there already is a command `!searchqupte`. But I'd argue it serves a different purpose. It'll return all quote ids that contain the keyword. So it's a slightly different use case.

BrandenB (approved):

gmt2001 (approved):

PR #2332, opened 2020-07-21
utf8mb4 charset for MySQL

emilweth:
Hi,

When using MySQL datastore, if I use emojis in a command, the command isn't saved because the default charset can't handle it.

See 
image

But it's working with utf8mb4 

image

And it's correctly reflected in phantombot interface

image

Also, if the database has already been created, it must be updated

```
ALTER DATABASE phantombot CHARACTER SET = utf8mb4 COLLATE = utf8mb4_unicode_ci;
```

```
ALTER TABLE phantombot_<table> CONVERT TO CHARACTER SET utf8mb4 COLLATE utf8mb4_unicode_ci;
```

If the database does not exists yet, it must be created with

```
CREATE DATABASE phantombot COLLATE 'utf8mb4_unicode_ci';
```

_(please don't judge my obnoxious usage of emojis)_

### DIFF
--- a/source/com/gmt2001/datastore/MySQLStore.java
+++ b/source/com/gmt2001/datastore/MySQLStore.java
@@ -128,7 +128,7 @@ public class MySQLStore extends DataStore {
         fName = validateFname(fName);
 
         try (Statement statement = connection.createStatement()) {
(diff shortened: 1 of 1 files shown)


gmt2001:
I feel like I fixed the not using utf8mb4 issue previously, but it never got merged. No idea why

gmt2001 (approved):

PR #2362, opened 2020-08-28
Fix The choice of Java method java.lang.String.replace matching JavaS…

gmt2001:
…cript argument types (function,string) is ambiguous


gmt2001:
closes PhantomBotDE/PhantomBotDE#33

PR #2397, opened 2020-10-08
Remove getModuleIndex()

Arb-aya:
**javascript-source/init.js**
As modules are added as properties of the array object the array has no length, nor can indexOf find any elements and thus will always return -1.

* Removed the getModuleIndex() function


BrandenB:
Please also remove the export of the function to the global $ API

BrandenB (changes requested):
Remove export to global $ API.

Arb-aya:
Good spot, done.

gmt2001 (approved):

PR #2410, opened 2020-11-02
Escape keyword when auto expanding keywordcount

robsdedude:
When keywords contain parentheses (quite common for regex keywords), putting `(keywordcount)` into the response will be expanded such that the keyword is expanded like a tag again. In the worst case, the users causes unexpected side effects. If not that, the bot will complain that the keyword is unknown.

### DIFF
--- a/javascript-source/core/commandTags.js
+++ b/javascript-source/core/commandTags.js
@@ -28,8 +28,9 @@
 
     /*
      * @function unescapeTags
-     *
+     * @export $
      * @param {string} args
+     * @returns {string}
      */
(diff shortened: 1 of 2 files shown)


javascript-source/core/commandTags.js
```diff
@@ -1698,5 +1698,6 @@
     }
     
     $.tags = tags;
+    $.escapeTags = escapeTags;
```
robsdedude, line 1711: Add export info to function's doc comment.

gmt2001 (approved):

PR #2412, opened 2020-11-03
Add external and hidden commands + add panel options for those and disabled commands

robsdedude:
Phantombot already supports disabling commands via `!disablecom` and `!enablecom`. This feature is now accessible through the web panels as well. Furthermore I added the option to hide commands so that they don't get listed with `!commands`. This helps to reduce clutter. Finally, I've added external commands. Those are commands that the bot does not handle but list under `!commands`. This can be helpful if you have other bots in chat but you don't want them to respond to the `!commands` command too. Workarounds for this feature already exist: either add a custom command with empty response or adjust the translation of the `!commands` command. Both seem like unclean hacks to me.

todo:
 - [x] hide and disable aliases

### EDIT:
Here are some screenshots.  
The updated custom commands page:
image

Edit command modal:
image

Edit command modal with extended advanced section:
image

The updated alias command page:
image

Edit alias modal (has no advance section)
image

The new external commands page:
image

External command modal:
image


javascript-source/commands/customCommands.js
```diff
@@ -733,6 +741,109 @@
             return;
         }
 
+        /*
+         * @commandpath hide [command] - Hide a command in the !commands list
```
gmt2001: Comment has wrong command name

gmt2001 (dismissed):

BrandenB (changes requested):
I personally don't want to see external commands under the custom command section, it should go under commands > external. No where else. Otherwize you end-up with a bunch of stuff under one tab that isn't related. 

There also needs to be a loading circle under the external commands like custom commands.

Also, under advanced, when adding or editing a command there should be a checkbox with the option to hide it from the commands list.

Also follow the panel standard JS code, don't create a bunch of functions everywhere. Functions should also be declared as 

```const functionName = function() {```

If the panel cannot do what all commands do it won't get merged.

Please also upload pictures of the final product on the panel.

Thanks.

robsdedude:
> I personally don't want to see external commands under the custom command section, it should go under commands > external. No where else. Otherwize you end-up with a bunch of stuff under one tab that isn't related.

Makes sense. I put it there initially because I thought if it gets disabled together with the custom commands module it makes sense to have it on the same page. But same applies to aliases and it's its own page too. I'll make it so.


> Also, under advanced, when adding or editing a command there should be a checkbox with the option to hide it from the commands list.

I am confused. I thought this is exactly what I've done?

> Also follow the panel standard JS code, don't create a bunch of functions everywhere. Functions should also be declared as
> `const functionName = function() {`

I'm not sure I understood that correctly. Adding helper functions is ok, when I assign them to a const like above or not?

> If the panel cannot do what all commands do it won't get merged.

Funfact: disabling and enabling commands is currently (latest stable) possible only via commands. This PR aims exactly at fixing that + adding more features.

> Thanks.

Thanks for the fast and in-depth review.

robsdedude:
Added features for aliases, added some screenshots, split external commands to own page, and updated code style.

BrandenB (approved):
Looks good now, will let gmt go over it once more.


Thanks!

gmt2001 (approved):

robsdedude:
I just realized, that default commands (e.g. `!mods`) can be disabled via chat command `!disablecom !mods` as well. Still there is no way to do so via the web panel. Probably worth another PR.

BrandenB:
If you want to do it, go ahead we'll merge it.

gmt2001:
closes gmt2001/PhantomBot-1#120

PR #2417, opened 2020-11-07
WIP: Welcome System

robsdedude:
This adds a new alert that will make the bot greet new chatters after a time out. Optionally, a special message can be chosen for first time chatters.

This is majorly different compared to the greeting system in that it
 - is a system where the bot greets a chatter and not a user all other chatters,
 - is based on people interacting with chat and not the IRC join event,
 - and the same message is used for everyone (because of the first point)

image

This PR is WIP because I didn't have the chance to test it. I'll update this PR when I have used the system for a while.

### DIFF
--- a/javascript-source/systems/welcomeSystem.js
+++ b/javascript-source/systems/welcomeSystem.js
@@ -1,0 +1,218 @@
+/*
+ * Copyright (C) 2016-2020 phantombot.github.io/PhantomBot
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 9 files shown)


BrandenB (dismissed):
I like this, however ever maybe for the excluded users we should do the same as we do for Discord command permissions (without the dropdown list of course).

Anyway, regardless I approve.

javascript-source/systems/welcomeSystem.js
```diff
@@ -0,0 +1,240 @@
...
+            if (!welcomeQueue.isEmpty() && welcomeEnabled) {
+                $.say(welcomeQueue.poll());
+            }
+
+            /* There are welcomes, however, welcome has been disabled, so destroy the queue. */
+            if (!welcomeQueue.isEmpty() && !welcomeEnabled) {
+                welcomeQueue = new java.util.concurrent.ConcurrentLinkedQueue;
+            }
+
+        }, 1000, 'scripts::systems::welcomeSystem.js');
```
BrandenB: Make this longer, 1 second will spam the chat if a streamer gets raided. Or make it batch users together.

`user1, user2, user3 joined the channel`

BrandenB (approved):
Perfect, thanks a lot.

PR #2431, opened 2021-01-01
Patch 235

### LINKED ISSUE #2425: Phantom don't react from autohost events
The bot react to normal host events but I don't see any autohost event reaction and it has an implementation in the code.

gmt2001:


gmt2001 (commented):
closes PhantomBot/PhantomBot#2425

PR #2447, opened 2021-02-05
Add round for "(hours)" variable

smeinecke:
Unfortunately, the variable "(hours)" was output without round, which leads to it being output with many decimal places.

Example:
admin: !addcom !chattime (@sender) has been chatting for (hours) hours!

user: !chattime 
bot: @user has been chatting for 30.8333333333332 hours!

*after change*
bot: @user has been chatting for 30.8 hours!

### DIFF
--- a/javascript-source/core/commandTags.js
+++ b/javascript-source/core/commandTags.js
@@ -739,7 +739,7 @@
                     user = String(event.getSender());
                 }
                 return {
-                    result: String($.getUserTime(user) / 3600),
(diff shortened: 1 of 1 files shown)


gmt2001 (dismissed):

gmt2001 (approved):

gmt2001:
Approved with a change

Moved your new implementation to `(hoursround)` just to avoid the inevitable complaining from users who don't like change

PR #2528, opened 2021-07-26
Add multi timer support

robsdedude:
Proposing a new feature: having multiple timer/notice groups. Each having their own settings and timer running.
This also adds the option to shuffle the messages in groups and send them at random times (upper and lower limit needs to be specified; uniformly distributed).

I'll mark this as WIP because I only tested everything briefly. I'll deploy it and remove the WIP state when I've successfully used it for a while. Feedback is very welcome already.

It looks like this:
image
or on a smaller screen:
image


robsdedude:
I *think* this is ready to be merged now. I've used it for quite a while now and so far my bot did not catch fire. **But** this is a rather big change (+1337 lines, `nice`). Please test this feature yourself again before merging it.

javascript-source/core/updates.js
```diff
@@ -228,6 +229,66 @@
         $.inidb.set('updates', 'installedv3.5.0', 'true');
     }
 
+    /* version 3.4.7 updates */
```
robsdedude, line 232: a) I just realized the version in the comment is not what the following code says.
b) Will this feature go into 3.5 anyway?

gmt2001: Yes

The update should still go off anyway, but if you want you can move it to the 3.5.0 block

gmt2001 (approved):

gmt2001:
closes gmt2001/PhantomBot-1#148

robsdedude:
:partying_face:

PR #2553, opened 2021-09-08
Allow 3.4.7 to 3.5.0 Upgrade to succeed

ZeldaZach:
While upgrading from PhantomBot 3.4.7 to 3.5.0 nightly, a fatal exception can be thrown if a null value is found.

"TypeError: cannot call method "startsWith" of null (core/updates.js:222)"

This was handled by ensuring the value is well established before checking the string for content.


ZeldaZach:
Bump on this one :)

PR #2713, opened 2022-03-27
Fix #2583

### LINKED ISSUE #2583: Bot resets game
PhantomBot Version: 3.5.0-NB-11072021 (nightly_build) (Revision: 53cb050)
Operating System: Linux
Java Version: 11.0.4+10-LTS
Web Browser Version (if Panel related):

Detailed Description: When I start streaming, the bot usually posts on Twitter with the wrong game and later on posts a "game changed to"-tweet.
Sometimes, however, the bot resets the game to the last one played (e.g. was "just chatting" and will be changed to "Guild Wars 2"). Maybe the bot could pull the game from twitch when detecting that a stream is online and posts after updating itself?
I really don't know what causes the bug. If I set the game a certain time beforehand, this does not happen.
I've tried to set the game the day before -> works
I've tried a few minutes beforehand -> doesn't work

Steps to Reproduce:

Logs:

Screenshots (if applicable):

Sartharon:
Fixes [#2583](https://github.com/PhantomBot/PhantomBot/issues/2583)

The issue is that EventSub sends an Online Event which triggers twitter to post a tweat while at the same time retrieving the game from TwitchCache. But The cache still hold the old game. Later on, the cache gets updated and the game changes.
Though this is not the best solution imo but it fixes the issue by forcing the update first and then triggers the "online"-Event. Thus, at the time the twitter script polls the cache for the game it's already updated.

P.s Sorry for the preliminary Pull req before I added this message ... my bad I miss clicked

Edit: Sadly there is no game information provided by PubSub [only this](https://ibb.co/pdW0whz) which is why I use the updateCache function

### DIFF
--- a/source/tv/phantombot/cache/TwitchCache.java
+++ b/source/tv/phantombot/cache/TwitchCache.java
@@ -32,6 +32,8 @@ import java.util.Date;
 import java.util.Map;
 import java.util.TimeZone;
 import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.locks.ReentrantLock;
+
 import org.apache.commons.io.FileUtils;
 import org.json.JSONArray;
 import org.json.JSONException;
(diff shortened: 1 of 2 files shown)


Sartharon:
Another Idea would be to write a separate "Update-Game" function, but this only causes duplicate code. Considering that online and offline events should only happen once per stream, running updateCache twice at these times shouldn't pose an issue.

Though I'm open for suggestions and open to implement them

gmt2001 (approved):

Issue #2717, opened 2022-03-29
!shoutout command does not use the 'corecommands.shoutout.no.game' response if the target does not have a recent game.


hippysec:
PhantomBot Version: 3.5.2.3
O/S Version: Debian 11, Kernel 5.10.0-12-amd64

Java Version:
openjdk version "11.0.14.1" 2022-02-08
OpenJDK Runtime Environment Temurin-11.0.14.1+1
OpenJDK 64-bit Server VM Temurin-11.0.14.1+1

Web Browser Version (if Panel related):NA

Detailed Description:
When using the !shoutout command against a twitch username that has not streamed, it does not trigger the IF logic to trigger the 'corecommands.shoutout.no.game' response and instead moves to the IF for !$.isOnline. and responds in chat with the 'corecommands.shoutout.offline' response.

I will be submitting a pull request on coreCommands.js with a proposed resolution that I've successfully tested.

Steps to Reproduce:
Use the !shoutout command against a Twitch username that has not streamed any games. 

I'm only really comfortable in Powershell right now, but if there is a good way to pull out any additional logging or information I would be happy to provide it with some direction on how to obtain it.

Logs:
From the journalctl output (I used a public Twitch bot with no stream history as an example. hopefully that is acceptable):
Mar 27 22:41:02 PhantomBotVM0 launch-service.sh[27391]: p03-28-2022 @ 02:41:02.912 GMT] hippysec: !shoutout Zeldobot
Mar 27 22:41:03 PhantomBotVM0 launch-service.sh[27391]: p03-28-2022 @ 02:41:03.912 GMT] [CHAT] Shout out to ZeldoBot! They were recently playing  @ https://twitch.tv/zeldobot! Check them out and consider following. :)

=====

excerpt from my core-coreCommands.js as I have modified it to output a different message than the default. :
$.lang.register('corecommands.shoutout.offline', 'Shout out to $1! They were recently playing $3 @ $2! Check them out and consider following. :)');
$.lang.register('corecommands.shoutout.online','Shout out to $1! They are playing $3 RIGHT NOW! Check them out and consider following. :)');
$.lang.register('corecommands.shoutout.no.game','Shout out to $1! They have not played anything recently, but consider following @ $2. :)');

PR #2718, opened 2022-03-29
Fix Issue #2717, !shoutout not triggering corecommands.shoutout.no.game message.

hippysec:
#2717 

When using !shoutout against a "streamer" (aka. Twitch User) that has not streamed the "corecommands.shoutout.offline" response is being triggered and sent in chat. The logic on the original line 42 did not trigger due to whatever value was captured in variable "streamerGame". I added an additional comparison. With the added comparison (streamerGame == ""), it is now using the appropriate "corecommands.shoutout.no.game" response when shouting out a streamer/Twitch User that has not streamed.

PhantomBot/Javascript-Source/Core/coreCommands.js
Line 42 originally reads as:
if (streamerGame == null || streamerGame.length === 0) {

The proposed new line 42 reads as:
if (streamerGame == null || streamerGame.length === 0 || streamerGame == "") {

since I forgot to add the a log output showing the post-change results that I tested in the original issue (2717), please see the log entries from my journalctl output below - which shows a chat message with the appropriate output for a user that has not streamed.
Mar 27 23:18:36 PhantomBotVM0 launch-service.sh [28612]: [03-28-2022] hippysec !shoutout Zeldobot
Mar 27 23:18:37 PhantomBotVM0 launch-service.sh [28612]: [03-28-2022] [CHAT] Shout out to ZeldoBot! They have not played anything recently, but consider following @ https://twitch.tv/zeldobot. :)

### DIFF
--- a/javascript-source/core/coreCommands.js
+++ b/javascript-source/core/coreCommands.js
@@ -39,7 +39,7 @@
                     streamerGame = $.getGame(streamer),
                     streamerURL = 'https://twitch.tv/' + streamer;
 
-            if (streamerGame == null || streamerGame.length === 0) {
+            if (streamerGame == null || streamerGame.length === 0 || streamerGame == "") {
                 $.say($.lang.get('corecommands.shoutout.no.game', streamerDisplay, streamerURL));
                 return;
             }


gmt2001:
Thank you for your contribution. After looking at the original code, I have determined that there was a more appropriate fix for this, which has been tagged and will auto-close this PR.

PR #2720, opened 2022-03-29
Separate global and per-user command cooldown

Sartharon:
Separates the per command cooldowns into global (cooldowns for everyone) and per-user cooldowns, as mentioned in [this issue](https://github.com/gmt2001/PhantomBot-1/issues/8).
The bot-wide global cooldown is used, if none is set for a specific command.
Thus, this PR gives the user more configurability on how cooldowns work, such as only allowing a user to use a command once an hour while others can still use it freely (as far as the global cooldown allows it).
Further, this PR improves the code flow of **commandCoolDown.js** and simplifying it as well. (for example, the `getSec()` function was only used in `init.js` and while being **95%** identical to the `get()` function.)

This PR introduces changes to the `!coolcom [command] [seconds] [type (global / user)]` command. The new syntax is as follows:
+ `!coolcom [command] [user=seconds] [global=seconds]` where user and global can be used interchangeably and also allowing a single number.
+ `!coolcom age user=130 global=30` //Sets per-user cooldown to 130 seconds and the global cooldown to 30 seconds
_is equal to_
 `!coolcom age global=30 user=130`
+ `!coolcom age global=30` //Sets only the global cooldown to 30 seconds and leaves the per-user cooldown unchanged
_is equal to_
 `!coolcom age 30` //Assumes global, if no type (user/global) is given and thus setting the global cooldown to 30 seconds
+ `!coolcom age user=130` //Sets only the user cooldown to 130 seconds and leaves the global cooldown unchanged

To allow for this separation, an update to the database is necessary, which will use the following new layout:
_Command_Name - JSON {command: Command_Name, globalSec: Global_Cooldown, userSec: User_Cooldown}_
Databases created before this change is released will be upgraded accordingly.

The Panel had to reflect this change as well to allow users to also configure per user commands for audio hooks, custom commands and default commands there. This is achieved by replacing the **Global**-checkbox with an input field of type **number**:
image

The related lang-files have been modified as well.

An example user interaction would be this:
image

P.s. I assumed this will be part of 3.6.0 which is why that's the version being saved in the database during the update.

### DIFF
--- a/javascript-source/core/commandCoolDown.js
+++ b/javascript-source/core/commandCoolDown.js
@@ -31,19 +31,41 @@
 
     $.raffleCommand = null;
 
+    const   Type = {
+                User: 'user',
+                Global: 'global',
+            },
+            Operation = {
+                UnSet: -1,
+                UnChanged: 0,
+            };
+
     /*
      * @class Cooldown
-     *
      * @param {String}  command
-     * @param {Number}  seconds
-     * @param {Boolean} isGlobal
+     * @param {Number}  globalSec
+     * @param {Number}  userSec
      */
-    function Cooldown(command, seconds, isGlobal) {
-        this.isGlobal = isGlobal;
-        this.command = command;
-        this.seconds = seconds;
-        this.cooldowns = [];
-        this.time = 0;
+    function Cooldown(command, globalSec, userSec) {
+            this.command = command;
+            this.globalSec = globalSec;
+            this.userSec = userSec;
+            this.userTimes = [];
+            this.globalTime = 0; 
+    }
+
+    /*
+     * @function Stringify
+     * @param {String}  command
+     * 
+     * @return {JSON String}
+     */ 
+    function toJSONString(command) {
+        return JSON.stringify({
+            command: String(command.command),
+            globalSec: command.globalSec,
+            userSec: command.userSec
+        });
     }
 
     /*
@@ -56,8 +78,7 @@
 
         for (i in commands) {
             json = JSON.parse($.inidb.get('cooldown', commands[i]));
-
-            cooldowns[commands[i]] = new Cooldown(json.command, json.seconds, json.isGlobal.toString().equals('true'));
+            cooldowns[commands[i]] = new Cooldown(json.command, json.globalSec, json.userSec);
         }
     }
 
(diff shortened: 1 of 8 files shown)


javascript-source/init.js
```diff
@@ -512,34 +512,44 @@
...
+            } 
+
+            [cooldownDuration, isGlobalCooldown] = $.coolDown.get(cooldownCommand, sender, isMod);
+
+            if (cooldownDuration > 0) {
+                if (isGlobalCooldown) {
+                    $.sayWithTimeout($.whisperPrefix(sender) + $.lang.get('init.cooldown.msg.global', command, cooldownDuration), $.getIniDbBoolean('settings', 'coolDownMsgEnabled', false));
+                    consoleDebug('Command ! ' + command + ' was not sent due to it being on a global cooldown.');
+                } else {
+                    $.say($.whisperPrefix(sender) + $.lang.get('init.cooldown.msg.user', command, cooldownDuration), $.getIniDbBoolean('settings', 'coolDownMsgEnabled', false));
```
gmt2001: $.say does not have a second parameter

$.sayWithTimeout blocks the bot from sending the message if $.sayWithTimeout has sent a message within the last 10 seconds

The second parameter in $.sayWithTimeout was added by someone who was lazy to replace a normal `if (thisOutputIsEnabled)`

Either replace $.say with $.sayWithTimeout or remove the second parameter and just do a normal if block

javascript-source/core/updates.js
```diff
@@ -295,4 +295,40 @@
...
+
+            $.inidb.set('cooldown', cooldowns[i], JSON.stringify({
+                command: String(json.command),
+                globalSec: globalSec,
+                userSec: userSec
+            }));
+        }
+
+        $.consoleLn('PhantomBot update 3.6.0 completed!');
+        $.inidb.SetBoolean('updates', '', 'installedv3.6.0', 'true');
```
gmt2001: As this is being set as a Boolean in line with my recent patch into nightly, please submit the actual boolean value `true` instead of a string

javascript-source/core/commandCoolDown.js
```diff
@@ -95,76 +116,69 @@
...
+        }
 
         if (isSpecial(command)) {
-            if (command == 'adventure' && defaultCooldowns[command] !== undefined && defaultCooldowns[command] > $.systemTime()) {
-                return defaultCooldowns[command];
-            } else {
-                return 0;
+            if (command.equalsIgnoreCase('adventure') && defaultCooldowns[command] !== undefined && defaultCooldowns[command] > $.systemTime()) {
+                maxCoolDown = getTimeDif(defaultCooldowns[command]);
+                return [maxCoolDown, isGlobal];
```
gmt2001: Return is redundant

gmt2001 (changes requested):
This looks really good. There are just a few minor changes requested

javascript-source/init.js
Sartharon: I left the `$.sayWithTimeout` on purpose in the case of a global cooldown due to the fact that the bot can potentially get spammy:
User 1: !age
Bot: Standard output message
User 2: !age
Bot: Global cooldown message
User 3: !age
Bot: Global cooldown message

Using `$.sayWithTimeout` elevates this potential spam problem. Though if you prefer to also just use `$.say` here I'll change it.

On the other hand I will still need to check if the "Send Cooldown Message" option is enabled, currently the per-user cooldown message is sent even without the option not being turned on.

javascript-source/core/commandCoolDown.js
Sartharon: Thanks ... a little oversight from a last minute cleanup I've made to improve the codeflow. I will commit these in a few hours

gmt2001 (approved):

Issue #2627, opened 2021-12-15
[Feature Request] Automatically distribute points to Raffle Winner


priestkos:
PhantomBot Version: 3.5.0.4
O/S Version: Windows 10 21H2
Java Version: Version 8 Update 311

Detailed Description: 
About Raffles and Ticket Raffles: 
1. Ability to draw more than one winner 
2. An option to automatically distribute loyalty points to the aforementioned winner/winners

Thanks for the work you put in!

PR #2735, opened 2022-04-04
[TicketRaffle] Improve Ticket Raffle

Sartharon:
**Original PR message:**
_This PR aims to get in line with other bots as mentioned on discord.

This PR enables the user to buy the maximum amount of tickets permitted via the use of `!tickets max` while keeping the amount of tickets the user already bought and is able to buy with their points in mind._

**Edit:**
Additionally, to the original PRs purpose, this PR also aims to incorporate the requested ticket raffle changes from [this issue](https://github.com/PhantomBot/PhantomBot/issues/2627). Mainly:

1. The ability to draw multiple winners at the same time with
2. The ability to distribute the bots' loyalty points to winners

To achieve 1. and 2. the `!draw` command was expanded to accept the following syntax `!draw [amount] [prize]`. Where amount is the amount of winners to be drawn and defaults to 1 if not set, prize is the amount of points each winner will be awarded with and defaults to 0 if not set.
Examples:

- `!draw 10` - Draws 10 winners if enough unique players have entered
- `!draw` - Draws a single winner
- `!draw 2` 1000 - Draws 2 winners and awards 1000 loyalty points to each winner

image

In order to allow for large amounts of winners to be drawn, a faster 'pick random users' method is used which also checks if the drawn user from the entries has not yet been drawn. The winners are announced in chat, due to the message limit of 500 characters a check is in place to splice the message in two separate messages allowing for greater winner amounts.

Before this change redrawing was possible allowing to draw multiple winners, though this was prone to selecting the same player multiple times. In all fairness, with the ability to draw multiple winners at once, I don't see the need to redraw winners and have thus blocked the command if winners were already drawn. Though, I'm open for a discussion or suggestions on this matter and will happily implement a !redraw command or just expand on !draw. Considerations here would be to deduct the loyalty points awarded before.

These changes are reflected in the panel as well:

- The Draw button will now be disabled if winners were already drawn
- Draw Options were added to account for the new parameters accepted by the expanded `!draw` command
image

- The winners will now be shown in the table/list once the ticket raffle has ended and winners were drawn to allow the user/streamer to read the winners from the panel and not having to look at the chat which scrolls the winners away
image

### DIFF
--- a/javascript-source/lang/english/systems/systems-ticketraffleSystem.js
+++ b/javascript-source/lang/english/systems/systems-ticketraffleSystem.js
@@ -32,8 +32,8 @@ $.lang.register('ticketrafflesystem.entered.bonus', '$1 (+ $2 bonus) entries add
 $.lang.register('ticketrafflesystem.usage', 'Usage: !traffle open [max entries] [regulars ticket multiplier (default = 1)] [subscribers ticket multiplier (default = 1)] [cost] [-followers]');
 $.lang.register('ticketrafflesystem.msg.enabled', 'Ticket raffle message\'s have been enabled.');
 $.lang.register('ticketrafflesystem.msg.disabled', 'Ticket raffle message\'s have been disabled.');
-$.lang.register('ticketrafflesystem.ticket.usage', 'Usage: !tickets (amount) - And you currently have $1 tickets.');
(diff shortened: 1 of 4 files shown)


Sartharon:
I'll start the work on the "normal" raffle system to fulfill the issue mentioned in the PRs description when this gets approved

gmt2001 (dismissed):

gmt2001 (approved):

gmt2001 (approved):

PR #2770, opened 2022-04-23
check if user is root in the entry script

codebam:


codebam:
Closed because upstream doesn't do user checks in the entryfile anymore

PR #2898, opened 2022-08-19
Update streamHandler.js

zpammer:
No clue what the reason is for chopping off the name of the game after just 15 characters.
I took 45, but that is arbitrary as well.

I can imagine some UI issues on some devices occured in the past, but other content should get sliced than as well.
This works great on my 5 year old Android phone. 45 characters doesnt reach the end of a line in discord.

**Testresults**
I already use 45 for a long time; digged up an old post of the bot and one of today with the current nightly.

image
image


gmt2001:
@BrandenB did it

gmt2001 (approved):

PR #2906, opened 2022-08-22
Fix: Autolevel

Sartharon:
- Fixes user autoleveling.
- Ignores Twitch bots (they do not need to be leveled since the user added them to the bot list and wants them ignored anyway)
- Halves required lookup queries for each user check in the loop and should thus reduce server load

### DIFF
--- a/javascript-source/core/permissions.js
+++ b/javascript-source/core/permissions.js
@@ -359,6 +359,26 @@
     }
 
     /**
+     * @function isViewer
+     * @export $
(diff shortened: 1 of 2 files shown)


Sartharon:
Tbh I'm not sure if the secondary check is still nessecary

Edit: or one could modify `isMod()` to also check oMode and the mod array 🤔

gmt2001 (approved):

PR #2938, opened 2022-10-02
Band-aid the noticeSystem

Sartharon:
Adds a missing check for undefined in the notice-system to prevent:

`[ERROR] [init.js:357] Error with Event Handler [initReady] Script [./systems/noticeSystem.js] Stacktrace [sendNotice()@noticeSystem.js:273 > trySendNotice()@noticeSystem.js:194 > startNoticeTimer()@noticeSystem.js:130 > startNoticeTimers()@noticeSystem.js:103 > noticeSystem.js:865 > init.js:349 > init.js:480] Exception [TypeError: Cannot call method "startsWith" of undefined]`

### DIFF
--- a/javascript-source/systems/noticeSystem.js
+++ b/javascript-source/systems/noticeSystem.js
@@ -262,7 +262,7 @@
             }
         }
 
-        if (notice === null) {
+        if (notice === null || notice === undefined) {
             noticeLock.unlock();
             return false;
         }
--- a/javascript-source/systems/noticeSystem.js
+++ b/javascript-source/systems/noticeSystem.js
@@ -111,7 +111,8 @@
     /**
      * @function startNoticeTimer
      */
-    function startNoticeTimer(idx, retryCall = false) {
+    function startNoticeTimer(idx, retryCall) {
+        retryCall = (retryCall === undefined || retryCall === null) ? false : retryCall;
         noticeLock.lock();
         stopNoticeTimer(idx);
 
--- a/javascript-source/systems/noticeSystem.js
+++ b/javascript-source/systems/noticeSystem.js
@@ -63,11 +63,11 @@
             intervalMin = noticeGroups[noticeGroups.length - 1].intervalMin;
             intervalMax = noticeGroups[noticeGroups.length - 1].intervalMax;
 
-            if (intervalMin === null) {
+            if (intervalMin === null || intervalMin === undefined) {
                 intervalMin = intervalMax || 10;
                 inconsistent = true;
             }
-            if (intervalMax === null) {
+            if (intervalMax === null || intervalMax === undefined) {
                 intervalMax = intervalMin;
                 inconsistent = true;
             }
@@ -112,11 +112,7 @@
     /**
      * @function startNoticeTimer
      */
-    function startNoticeTimer(idx, retryCall) {
-        if (retryCall === null) {
-            retryCall = false;
-        }
-
+    function startNoticeTimer(idx, retryCall = false) {
         noticeLock.lock();
         stopNoticeTimer(idx);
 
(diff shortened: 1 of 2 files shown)


Sartharon:
Alright, I suppose I've got the error sorted properly now.
Did some code formatting as well (parts were so unreadable).

The locks were all over the place ... still, I do not see why those would even be needed in a module, but I digress.
Ready to be merged imo.

P.s I've added this to the top of my cleanup/rework todo-list (especially getting rid of those nasty and useless locks ... but that would need to get properly tested before merging, for which I do not have time right now)

javascript-source/systems/noticeSystem.js
```diff
@@ -107,10 +112,7 @@
     /**
      * @function startNoticeTimer
      */
-    function startNoticeTimer(idx, retryCall) {
-        if (retryCall === null) {
-            retryCall = false;
-        }
+    function startNoticeTimer(idx, retryCall = false) {
```
gmt2001: I'm pretty sure the base feature-set of Rhino does not support this style of setting the default value

javascript-source/systems/noticeSystem.js
```diff
@@ -120,18 +122,23 @@
...
+
         if (isNaN(time)) {
             time = 10 * 6e4;
         }
         if (isNaN(lastSent)) {
             lastSent = now - time;
         }
         if (!retryCall && lastSent + time <= now) {
+            noticeLock.unlock();
+
```
gmt2001, line 128: Locking multiple times from the same thread is safe on a ReentrantLock, as long as the same number of unlocks are performed. If another thread tries to lock, it will block until the holding thread has called unlock the correct number of times

What I'm noticing is actually missing for safety is try-finally or try-catch-finally blocks to ensure there is no possibility of the unlock failing to trigger. The correct way to do this is:
```
lockobj.lock();
try {
    //Do Something
} catch (e) {  //Optional: only if Exceptions or JS errors are expected
    //Handle exception or JS error
} finally {
    lockobj.unlock();
}
```

The finally block is also triggered if a return statement occurs anywhere inside the try block

gmt2001 (changes requested):

javascript-source/systems/noticeSystem.js
Sartharon: Right i forgot about that one!

javascript-source/systems/noticeSystem.js
Sartharon: Exactly which is why i'd like to cleanup the whole lock unlock mess and get rid of it.
I mostly just reordered the lock/unlock calls.

I'll scratch this PR and PR another bandaid for the initial issue. Another PR will follow with a cleaned up version of this if you approve with the mentioned rework since I dont see the point of adding hundreds of try chatches everywhere

gmt2001: At a minimum, the unlock at the beginning of each function should simply be followed by a try covering the entire method, followed by a finally with the unlock, and then there should be no unlocks in the middle of the functions

gmt2001: I don't know why I didn't catch this when the original PR was made for this system

Sartharon: All good

The stop-gap-solution seems good. I'll probably push another commit tomorrow with the requested changes.

A cleanup will follow soon(tm) then 😅

gmt2001: I'm not necessarily opposed to using the lock, to ensure we can't manipulate the array from multiple threads at the same time, since each EventBus call is run on a random thread from the ThreadPool

Sartharon: Still it doesnt seem necessary in other modules.
If thinking about threadsafety we'd need to rewrite a bunch of modules

gmt2001: We probably do need to rewrite some LUL

Sartharon: Thinking about concurenthashmaps or a synchronized collection which java provides should be usable here as well because of rhino.

That should reduce the need for manual locking/unlocking

javascript-source/systems/noticeSystem.js
Sartharon: Resolved in [bfcd02d](https://github.com/PhantomBot/PhantomBot/pull/2938/commits/bfcd02d6ab5077b143dadcbf1115067a341a2d5a)

gmt2001 (approved):

PR #3103, opened 2023-02-23
Update timer.js so first column to re-order

Psychoboy:
Previous iteration the user couldn't click the button due to it allowing to drag the whole row. This changes it so the first column (the ID) is what is clicked on to drag and change order


Psychoboy:
Closing as it doesn't actually resolve the issue

Psychoboy:
This doesn't resolve an issue where the new order doesn't actually commit and save, just visual thing and allows editing again. The actual saving appears to be broken. Tested with old code and new code and the order never actually commits.

gmt2001 (approved):

PR #3215, opened 2023-05-07
Added option to warn the chat when raiding to an offline channel

therentabrain:
You rarely want to raid to a channel that is not streaming.  The mobile app generally refuses to follow such a raid, leaving mobile users stuck on the raiding channel while everyone else goes nowhere good.  The web page will raid, and strand web twitch viewers in a dead chat.  

This is usually due to a typo or miscommunication.  PhantomBot can help avoid raid train breakdown!

I added a chat command to set what PB will say after the raiding message _if_ the recipient is currently not streaming.  You can then use /unraid and fix the situation before it is too late.

` !raid setoutgoingofflinemessage [message]`

If you never set one, behavior will be consistent (no offline warning).

I've been using this feature for a couple of months now and it works great for me.  It has saved some difficulty!  

```username: !raid lisabei
username used: /raid lisabei.
BotName: Raid now departing for lisabei!
BotName:  Custom Test Warning: That channel is currently offline! But I'm sure you know what you're doing.

### DIFF
--- a/javascript-source/handlers/raidHandler.js
+++ b/javascript-source/handlers/raidHandler.js
@@ -23,7 +23,7 @@
             raidReward = $.getSetIniDbNumber('raidSettings', 'raidReward', 0),
             raidOutMessage = $.getSetIniDbString('raidSettings', 'raidOutMessage', 'We are going to raid (username)! Go to their channel (url) now!'),
             raidOutSpam = $.getSetIniDbNumber('raidSettings', 'raidOutSpam', 1),
-            raidOutOfflineMessage = $.getSetIniDbString('raidSettings', 'raidOutOfflineMessage', 'Note: That channel is offline!  Use /unraid if you wish to cancel . . .');
+            raidOutOfflineMessage = $.getSetIniDbString('raidSettings', 'raidOutOfflineMessage', null);
 
     /*
      * @function Reloads the raid variables from the panel.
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: javascript-source/handlers/raidHandler.js [hunk 1]
```diff
@@ -35,6 +36,7 @@
         raidReward = $.getIniDbNumber('raidSettings', 'raidReward');
         raidOutMessage = $.getIniDbString('raidSettings', 'raidOutMessage');
         raidOutSpam = $.getIniDbNumber('raidSettings', 'raidOutSpam');
+        raidOutOfflineMessage = $.getIniDbNumber('raidSettings', 'raidOutOfflineMessage');
```
gmt2001: Casting string to number

gmt2001 (changes requested):

javascript-source/handlers/raidHandler.js
therentabrain: Oh strange, that must have resurfaced from an older draft!  I remember that happening and fixing it the same day.

therentabrain: Oh I see, the (nearly) same thing happens twice, and I did fix the first one but not the 2nd. :) Thanks! 

Do I need to somewhere accept your fix of my fix?

gmt2001: Just push the fixing commit to the same branch that the PR was opened on and I will re-review it

therentabrain: Did I github right?  I feel like I did several steps when I could have just done one or two just now, but there it is.

gmt2001 (approved):

PR #3223, opened 2023-05-11
add nixpkgs flake definition

kloenk:


gmt2001:
Sorry, I have decided that I do not want to add any package managers beyond what is already supported. You are welcome to maintain and submit a nixpkgs flake as a 3rd-party package under the terms of our GNU GPLv3 or later license

PR #3249, opened 2023-05-24
Fix potential race condition in UserNoticeTMIProcessor

Sartharon:
Fixes or at least minimizes potential race conditions*
Adds a new ant target to only build and copy the jar into the distribution folder

\* **This does not change the behavior for marking `fromBulk` if the `submysterygift`-Message is late**

### DIFF
--- a/build.xml
+++ b/build.xml
@@ -298,6 +298,23 @@
     <target name="check-compiled">
         <available file="${version.folder}/PhantomBot.jar" property="compiled"/>
     </target>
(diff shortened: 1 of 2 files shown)


Sartharon:
I missed a test scenario where massGiftSub and normal giftSubs occure in rapid and random succession.

LMK if I should test this. Will do that the coming day if requested

gmt2001 (approved):

PR #3321, opened 2023-07-22
Error logging in init.js, and fix for web panel, somehow combined into one PR - dunno why

therentabrain:
I put this line in my copy of init.js many moons ago.  I find it helpful for troubleshooting, and also it can be useful to see what disabled commands people try, because maybe they should be added.

Somehow it took three changes to make this right but oh well!


therentabrain:
Somehow my new change wound up part of this pull request.  I assumed I could make a new one.  um.  Ok.  Well there it is :)

gmt2001 (approved):

PR #3322, opened 2023-07-22
Change order to allow panel to open

therentabrain:
Per this thread:
https://discord.com/channels/107910097937682432/1132292565551566858 Panel would not load, claiming 
`helpers.getBranch is not a function`
I assume because it was not yet a function in the order of declarations. I moved helpers.canEmbedTwitch and helpers.CANT... down below where it was declared and it works now.

Disclaimer:  I am using http, not https.  It is possible that someone with https would get past this error, though I doubt it, because it's in the initial declarations.  Maybe those who aren't seeing this error are lucky enough to have the older js cached.


Sartharon:
You shouldnt edit files under /docs since those are generated by a workflow.
Edit the according file under /resources/web/

therentabrain:
I'm going to bed, but I sure don't mind if someone else wants to fix it while I am asleep :)

PR #3348, opened 2023-08-18
Fix null pointer on user creation

Sartharon:


Sartharon:
Closed due to #3349 being prefered

Issue #3466, opened 2023-12-09
!shoutoutapitoggle not functioning


IanDLive:
PhantomBot Version: 3.11.0.3 (Stable)
O/S Version: Windows Server 2019
Java Version: 17.0.8.1+1
Web Browser Version (if Panel related): N/A

Detailed Description:
After wanting to enable the Twitch API for the shoutout command I found that the command "!shoutoutapitoggle" isn't functioning, giving no response.

Steps to Reproduce:
Use of the command !shoutoutapitoggle

Logs:
After looking through the files, I found that on line 60 of ./scripts/core/coreCommands.js that the command that it is looking for in the IF statement is "shoutoutapi" not the registered command of "shoutoutapitoggle"

Changing "shoutoutapi" to "shoutoutapitoggle" on line 60 fixes the issue.

Screenshots (if applicable):

PR #3475, opened 2023-12-27
[Points/Time] Ensure tables exist

Sartharon:
Ensures tables are present when batch increases or sets are called. Otherwise, the time and points system will not complete their interval runs as expected (increase in time and points) on a fresh database.

No negative effects as there is no other usage for those functions besides from the permissions "visited" tracker and VIP/Sub swap


Sartharon:
Now also closes #3466 as the comparison didn't compare the input to the registered command

gmt2001 (approved):

javascript-source/core/coreCommands.js
gmt2001, line 1: closes #3466

PR #3547, opened 2024-10-18
Drastically speed up readfilerand transformer for large files

Sartharon:
As indicated on discord, the `readfilerand` transformer is rather inefficient as it relies on reading the whole file and passes it multiple times (as arguments) between multiple functions before finally returning a selected line.

This PR aims to speed up the selection by utilizing `RandomAccessFile` allowing a file to be randomly accessed

Example with test file with 1000000 lines (7.52MB) jumping throughout the whole range as expected

image

## Very rudimentary speed test

-  **Old Version:** +/- 5.8 s
image
- **New Version:** +/- 7 ns
image

### DIFF
--- a/source/com/gmt2001/JSFileSystem.java
+++ b/source/com/gmt2001/JSFileSystem.java
@@ -16,7 +16,9 @@
  */
 package com.gmt2001;
 
+import java.io.File;
 import java.io.IOException;
+import java.io.RandomAccessFile;
 import java.nio.file.FileAlreadyExistsException;
 import java.nio.file.FileVisitOption;
 import java.nio.file.Files;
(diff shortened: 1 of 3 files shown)


source/com/gmt2001/JSFileSystem.java
```diff
@@ -92,6 +95,40 @@ public static String ReadFileAsString(String path) throws IOException {
...
+     * @throws IOException If an I/O error occurs reading from the file or a malformed or unmappable byte sequence is read
+     */
+    public static String ReadFileRandLine(String path) throws IOException {
+        if (!PathValidator.isValidPathScript(path)) {
+            return null;
+        }
+
+        RandomAccessFile rFile = null;
+
+        try {
```
gmt2001: Prefer try-with-resources

gmt2001 (changes requested):

Sartharon:
Also added `0 byte` length (emtpy file) check to prevent `IOException` from being thrown when seeking empty files

gmt2001 (approved):

PR #3551, opened 2024-10-26
Fix (readfilerand) Umlaute

Sartharon:
Use byte encoding to get umlaut to work with `RandomFileAccess`

Before:
image

After:
image


Sartharon:
Might pose issues with characters not included in ISO-8859-1

PR #3666, opened 2026-01-12
Chore: full dependency upgrade

Sartharon:
### PR includes:

- **Migration to Rhino 1.9.0**
__
Includes dynamic compilation and linking drastically reducing the memory footprint (my tests show from +/-115 MB to easily below 90MB)
--
Rhino has been split into modules now. To facilitate first migration steps I resorted to import all for now. Later on we should be able to simply use `rhino` and `rhino-tools` further reducing the build footprint
--
To enable the afore mentioned dynamic linking we needed the `jdk.dynalink` therefore I've added it to the jlink pipeline. (Otherwise we may only run the Phantombot with `context.setOptimizationLevel()`; set to `0` or `-1` in any case optimization levels are deprecated. I am not aware what might break due to the dynamic linking but the bot starts and commands are working for me.
- **Migration to Discord4j 3.3.0**
__
`VoiceChannel` has been removed and been replace with `AudioChannel` everything was renamed accordingly
--
`ReactionEmoji` was replaced with `Emoji`
- **Migration to Netty 4.2.9 and reactor-netty 1.3.1**
__
reactors' `RetryRouter` was refactored to the new util.retry router
--
Netty's DNS resolver had to undergo a little refactor but will still try to use google DNS and revert to the OS provided resolver if needed (my firewall does block anything else than my own recursive DNS)
--
Netty 4.2.x is very strict when it comed to releasing reference counts and will throw exceptions if once releases something if the count is already 0. I've had to clean some manual release steps for this which might cause memory leaks. Though I ran the bot like this for a while with `-Dio.netty.leakDetection.level=PARANOID` but that is not a guarantee. I am happy for any contributions here as this is not as simple as the standard:
-- Sending will be auto released on `writeandflush()`
-- `SimpleChannelInboundHandler` extension handlers do not need to release any buffers if the super class isn't called with `false` as an parameter unless `retain()` is called
- **h2 upgrade path completed to 2.4.240**
- **Minor dependency upgrades**

### Memory profiling
This is by no means a scientific way of proving that there are no leaks even with the removal of some `ReferenceCount` releases but this with the `-Dio.netty.leakDetection.level=PARANOID` and spamming some requests+servings should indicate at least a general stability. (Please keep in mind this no active bot but a development environment)
### _**In an case I remain cautious in calling this "stable" and rather refer to this as an more than acceptable start**_
image
image
Context:
Initial heap peaks are loading the panel and memory initialisation of the bot itself
Slow rising heap peak is me spamming a plethora of alerts (50MB video and gif and with audio) with browser caching disabled.
After the alerts was transmitted we return backdown to a stable 75-80MB
AdaptiveByteBufAllocator is enabled by default in 4.2.x and tunes itself after a couple of requests which also might indicate the sawblade behavior at the start (see [their release notes](https://netty.io/news/2025/04/03/4-2-0.html) on this feature)
<br />
<br />

> My fight of optimizing Phantombot (maybe int oblivion 😄)
> To be continued ---- with webpage `script` and `CSS` imports


Sartharon:
The only `ByteBuf` we need to release are buffers we are the owner of:

1. https://github.com/PhantomBot/PhantomBot/blob/90b710333232837a3f766af85b630e339075a52d/source/com/gmt2001/httpwsserver/HttpServerPageHandler.java#L324
2. https://github.com/PhantomBot/PhantomBot/blob/90b710333232837a3f766af85b630e339075a52d/source/com/gmt2001/httpwsserver/HttpServerPageHandler.java#L326
3. https://github.com/PhantomBot/PhantomBot/blob/90b710333232837a3f766af85b630e339075a52d/source/com/gmt2001/httpwsserver/HttpServerPageHandler.java#L338
4. https://github.com/PhantomBot/PhantomBot/blob/90b710333232837a3f766af85b630e339075a52d/source/tv/phantombot/panel/WsPanelHandler.java#L572
5. https://github.com/PhantomBot/PhantomBot/blob/90b710333232837a3f766af85b630e339075a52d/source/tv/phantombot/panel/WsPanelHandler.java#L574

We don't need to touch any other `ByteBuf` as Phantombot uses `SimpleChannelInboundHandler` where Netty automatically starts the cleaning of these input buffers at the end of the pipeline (when calling `ctx.fireChannelRead()`).

For our writes into the channel we do not need to deal with any releases as this will be handled by the `HttpObjectEncoder` once the write has completed. (Anything else would anyhow not make sense as we'd release all references before the encoder could handle it)

Additionally, it seems the current request logger was not functional since a while (at least older versions did not logg either when I tried). Thus I took the liberty to also rewrite/fix it and make it option as to slim down the channel pipeline for each request.
Moreover I've moved the handler to fitting position in the pipeline so we can also logg any TLS request. Before only raw bytes could have been read as it was so early in the pipeline

Anything else should be pretty much similar to the previously expected behavior. A sample log looks as follows:

```
=== HTTP REQUEST === 2026-01-14_21:23:02.807672900 === HTTP REQUEST ===
Remote: 127.0.0.1:39139
Protocol: HTTP/1.1
Method: GET
URI: /panel/pages/loyalty/channelpoints.html?_=1768425692645
Headers:
  Host: <REDACTED>
  Connection: keep-alive
  Pragma: no-cache
  Cache-Control: no-cache
  sec-ch-ua-platform: "Windows"
  X-Requested-With: XMLHttpRequest
  User-Agent: Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/143.0.0.0 Safari/537.36 Edg/143.0.0.0
  Accept: text/html, */*; q=0.01
  sec-ch-ua: "Microsoft Edge";v="143", "Chromium";v="143", "Not A(Brand";v="24"
  DNT: 1
  sec-ch-ua-mobile: ?0
  Sec-Fetch-Site: same-origin
  Sec-Fetch-Mode: cors
  Sec-Fetch-Dest: empty
  Referer: <REDACTED>
  Accept-Encoding: gzip, deflate, br, zstd
  Accept-Language: en-GB,en;q=0.9,en-US;q=0.8,de-DE;q=0.7,de;q=0.6,es-ES;q=0.5,es;q=0.4
  Cookie: <REDACTED>
  content-length: 0
```


As for Head Space we stick close to the minimally required 70MB Phantombot needs to run

image

With a metaspace of about 100MB and all other stuff like Thread stacks I stay nicely under 300MB (on Windows) even with spamming the before mentioned big video files

image

gmt2001 (approved):

PR #3680, opened 2026-01-26
ConsoleInput overhead and a little threading

Sartharon:
### ConsoleInput
The `ConsolInputHandler` was constantly creating empty Strings thus imposing unnecessary activity on the heap. A 1,5 hour run showed **7%** of all the bytes allocated on the heap were coming from Thread-5 (in this case running the static `ConsoleInputHandler::run`)
image
image

While a static runner for this is fine I've give it a proper thread name as well now.
As are reading the console with `readlLine()` we are not required to sleep the thread as `readLine()` is already blocking until the line has completed either via `\n` or `\r` or by `EOF` there is no explicit need to sleep again.

One could think about move the ConsoleInputHandler as well on the ExecutorService. (I'm happy to follow up on this either via this PR or a new one if you'd like)

As I was already investigating the ConsoleInput I've taken the liberty to rearrange some of the InputEventHandlers flow allowing for earlier returns thus saving some branch checks

### About Threading in PhantomBot

As I've been looking at profilers and dumps over the last weeks and I noticed the thread naming scheme is a little hit or miss in PB. Thus this PR already starts by defining a default name for the ExecutionService's Threads

In general one can see some tasks/runners using the `ExecutionService` **and** setting a thread name. Setting a thread name is rather pointless as the thread will also run a plethora of other tasks.
As named threads only makes sense for static threads, my recommendation here is to remove this thread naming in itself for tasks running on the `ExecutionService` or at the very least revert it once the task has completed (mind you this comes with an overhead). The same is true for `JSTimer` which set thread names only valid briefly.

Let me know what you prever and I am happy to implement it

**Some minor thing**
I've also added some additional threadsafety for `JSTimer`

### DIFF
--- a/source/com/gmt2001/Console/in.java
+++ b/source/com/gmt2001/Console/in.java
@@ -44,12 +44,9 @@ public final class in {
 
             return s;
         } catch (IOException ex) {
-            try {
-                Thread.sleep(500);
-            } catch (InterruptedException ex2) {
-            }
+            com.gmt2001.Console.err.printStackTrace(ex);
         }
 
-        return "";
+        return null;
     }
 }
--- a/source/com/gmt2001/JSTimers.java
+++ b/source/com/gmt2001/JSTimers.java
@@ -21,6 +21,7 @@ import java.util.Objects;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ScheduledFuture;
 import java.util.concurrent.TimeUnit;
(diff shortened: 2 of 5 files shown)


gmt2001 (approved):

Sartharon:
I'm still open to your opinion regarding the thread naming topic and ready to act upon your preference unless you prefer to lay hand on it yourself 😎

PR #3683, opened 2026-02-01
Shared Rhino ContextFactory

Sartharon:
As discussed on discord, here an implementation for a shared context(factory). The reduction in loaded classes by this is negligible but it provides a cleaner code design

Additionally I' ve followed up on #3666 by only depending on the required rhino packages

### DIFF
--- a/source/tv/phantombot/PhantomBot.java
+++ b/source/tv/phantombot/PhantomBot.java
@@ -92,6 +92,7 @@ import tv.phantombot.httpserver.HttpSetupHandler;
 import tv.phantombot.panel.WsAlertsPollsHandler;
 import tv.phantombot.panel.WsPanelHandler;
 import tv.phantombot.panel.WsPanelRemoteLoginHandler;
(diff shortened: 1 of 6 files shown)


source/tv/phantombot/PhantomBot.java
```diff
@@ -973,6 +974,7 @@ private void initScripts() {
 
         /* And finally try to load init, that will then load the scripts */
         try {
+            RhinoRuntime.init(getEnableDebugging());
```
gmt2001: Rhino debugger should only enable if `getEnableRhinoDebugger()` is `true`

gmt2001 (changes requested):

source/tv/phantombot/PhantomBot.java
Sartharon: Good catch ... tab-/auto-complete got me on this one

gmt2001 (approved):

PR #3715, opened 2026-04-27
Fix missing donations & reduce unnecessary database calls

Sartharon:
- Fix donations being missed for two reasons
    - Greatest `donation_id` not necessarily representing the most recent donation -> base on `created_at` fallback to `donation_id`
    - `StreamLabsDonationEvent` only fired if `donation_id` is actually larger -> instead check each entry in the returned JSON. As we only request donations starting from the last processed one and limit the requested donations this is bound by `streamlabslimit` or `twitchalertslimit` or `5` thus being negligible in terms of used CPU cycles
    https://github.com/PhantomBot/PhantomBot/blob/e7df2fdd392515c556df22e8d299f8f2db8be14b/source/com/illusionaryone/StreamLabsAPI.java#L108
- Reduce constant database calls and saves of the `lastId` for a minimal increase in heap memory (worth the trade off at least in my eyes, we might even reduce the heap footprint by not creating JOOQ subordinate objects this way)

### DIFF
--- a/source/tv/phantombot/cache/DonationsCache.java
+++ b/source/tv/phantombot/cache/DonationsCache.java
@@ -47,6 +47,8 @@ public class DonationsCache implements Listener {
 
     private ScheduledFuture<?> updateFuture = null;
     private boolean firstUpdate = true;
+    private int lastId = PhantomBot.instance().getDataStore().GetInteger("settings", "", "DonationsCache_lastId");
+    private long lastTimestamp = 0;
 
     private DonationsCache() {
         this.updateFuture = ExecutorService.scheduleAtFixedRate(this::run, 20, 30, TimeUnit.SECONDS);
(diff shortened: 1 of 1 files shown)


Sartharon:
As an additional clarification:

This is an edge case and mainly occurs when multiple donations are recorded by StreamLabs at the same Unix Epoch (second based for them). If this occurs the outcome is rather undefined the main improvement here comes from moving the event firing block out of the `lastID` check due to the ordering of the events in the received JSON from their API
https://github.com/PhantomBot/PhantomBot/pull/3715/changes#diff-8721d0aecd00fdb59bf1f36e3315022e520780e97b87cfd055e2365b9b06edaaR96

The API states ["Results are ordered by creation date, descending."](https://dev.streamlabs.com/reference/donations)

During real streams we have observed returned objects like such (ordering matters here as well):

| donation_id |  created_at   |

|         22        | 1777709222 |
|         23        | 1777709222 |
|         21        | 1777709222 | 
|         24        | 1777709160 | 

With the current logic the 3rd donation would be lost and thus no event handlers fire. If the order of donations with id 22 and 23 would be switched we would even miss the 2nd and the 3rd. During peak times we observed up to 4 missed donations. 
After applying this patch we've not caught any such case even during donation spamming

gmt2001 (approved):

PR #3720, opened 2026-05-17
Feature: Let custom modules extend the panel via manifests without editing stock PB sources

mcawful:
## Brief description of the intended change

**Purpose:** Give **custom and community modules** a supported way to extend the **web panel**—**sidebar links** in the **`extra`**, **`alerts`**, **`giveaways`**, or **`audio`** submenus (via each **`nav`** item’s **`section`**), **Games** cards, and optional **settings / “more info”** UI—by shipping a **`manifest.json`** next to their usual Rhino scripts and panel pages, **instead of editing PhantomBot’s stock panel sources**. The bot **collects, validates, and serves** those manifests to the browser so installs (including **Docker** with a data volume) can pick up panel contributions safely after a refresh or reload.

---

## New files

**PhantomBot/source/com/mcawful/CustomPanelManifestCollector.java**

- Scan `web/panel/custom/**/manifest.json`, validate and merge `nav` / `cards`, warn on bad input — the panel must not trust raw manifests; this is the single server-side gate for schema, safety rules, deduplication, and merge order before any JSON is served.

**PhantomBot/source/com/mcawful/CustomPanelManifestCache.java**

- Memoize merged manifest JSON (and related cache metadata) for the panel endpoint — manifest scans can touch many files; caching keeps panel loads and polling cheap and enables **304 Not Modified** responses.

**PhantomBot/javascript-source/core/customScripts.js**

- Hot-scan `scripts/custom/` (and lang), `!reloadcustom`, and `initReady` replay for newly loaded scripts; invoked silently after manifest fetch from the panel — after a manifest refresh, new or updated Rhino modules should register commands without a full bot restart, and the panel can trigger that path safely.

**PhantomBot/resources/web/panel/js/utils/customPanelManifestLoader.js**

- Fetch `custom-manifests.json`, own `window.__pbCustomPanel__`, `initCustomPanelNav`, shared injected CSS (nav, modals, checkbox layout, community card flex row); `ensureStylesInjected` skips creating a second `#pb-custom-panel-styles` if one already exists — one place to load merged manifests, share constants/helpers across nav/cards/modals, inject small shared styles (including a stable **Games** card grid), and avoid duplicate `<style>` tags on SPA-style navigation.

**PhantomBot/resources/web/panel/js/utils/customPanelNav.js**

- Render manifest `nav` into the sidebar (dividers / sections) — stock `index.html` cannot list unknown modules; this injects links only into sidebar sections that already expose a submenu mount.

**PhantomBot/resources/web/panel/js/utils/customPanelCards.js**

- Render manifest `cards` on the Games page; toggles, settings, details; add `pb-custom-cards-mount` on the mount row — community cards need the same enable/disable and settings patterns (along with a new info pattern) as stock games without hand-authored HTML per module.

**PhantomBot/resources/web/panel/js/utils/customPanelSettingsModal.js**

- Declarative settings modal (flat fields or sections) and save path — modules should describe settings in JSON instead of shipping bespoke modal HTML/JS for every field shape we support.

**PhantomBot/resources/web/panel/js/utils/customPanelDetailsModal.js**

- Read-only details modal from manifest content — authors need a supported “about this module” surface tied to the same manifest as nav/cards.

**PhantomBot/docs/guides/content/developerdocs/custommodules.md**

- Developer doc for `manifest.json`, paths, and behavior overview — contributors need a canonical contract (paths, `section` values, field types) separate from scattered README fragments.

**PhantomBot/docs/guides/content/moduleguides/addingcustommodules.md**

- Installer-oriented module guide (links into developer docs) — installers follow the module guides tree first; this points them at manifest-based development when they outgrow “copy files only.”

---

## Modified files

**PhantomBot/source/tv/phantombot/httpserver/HTTPPanelAndYTHandler.java**

- Serve **GET `/panel/custom-manifests.json`** from the collector/cache (with **304** when unchanged) — the browser must load merged manifests through the same authenticated panel HTTP stack as the rest of the UI, with cache semantics.

**PhantomBot/source/com/gmt2001/PathValidator.java**

- Add `isPathUnderExecutionOrDockerWeb` (and related checks) so `./web` access is valid when custom content lives under the Docker data volume — Docker images symlink `web/panel/.../custom` into `PhantomBot_data`; checks that only allow paths under the JAR layout would wrongly **403** legitimate panel files.

**PhantomBot/source/tv/phantombot/httpserver/HttpSetupHandler.java**

- Use `PathValidator.isPathUnderExecutionOrDockerWeb` instead of `Reflect.GetExecutionPath()` prefix checks for `./web` auth paths — web-auth path validation must match where custom panel pages and assets actually live in Docker and bare-metal layouts.

**PhantomBot/source/tv/phantombot/httpserver/HTTPOAuthHandler.java**

- Same `PathValidator`-based web root check as `HttpSetupHandler` — OAuth-related static `./web` fetches use the same path rules; leaving one handler on the old check would be an inconsistent hole.

**PhantomBot/source/com/gmt2001/RollbarProvider.java**

- Include `com.mcawful` in `APP_PACKAGES` so stack traces from new classes are treated as first-party for Rollbar grouping — new collector/cache code lives under `com.mcawful`; without this, reports may be misclassified or filtered versus core PhantomBot packages.

**PhantomBot/resources/web/panel/index.html**

- Load `customPanel*.js` in dependency order — the loader must run before modals, nav, and cards; order is part of the public contract between those scripts.

**PhantomBot/resources/web/panel/js/index.js**

- Call `initCustomPanelNav()` after panel websocket auth — nav injection should run only when the panel session is real (socket ready), avoiding empty mounts or duplicate work on the login shell.

**PhantomBot/resources/web/panel/js/utils/ajaxLoader.js**

- Delegate `[data-folder]` clicks on `document` so dynamically injected custom nav links load pages — manifest nav `<a>` elements are added after the initial `$(function () { ... })` binding; direct binding on `[data-folder]` never sees them.

**PhantomBot/resources/web/panel/pages/games/games.html**

- Add `#pb-panel-games-custom-cards` mount and `injectCustomCards('games', …)` — stock Games markup has no mount point for third-party cards; this is the minimal hook that keeps community cards on the same page as built-in games.

**PhantomBot/resources/docker-entrypoint.sh**

- `mkdir -p` for `web/panel/pages/custom` and `web/panel/js/pages/custom` under `PhantomBot_data` — fresh data volumes may lack those directories; the bot should start clean even before the user copies any custom files.

**PhantomBot/Dockerfile**

- Stage/move/symlink `web/panel/custom`, `pages/custom`, and `js/pages/custom` into the data volume layout (alongside existing Docker steps) — custom panel files must persist on the volume and be visible under `/opt/PhantomBot/web/...` at runtime, matching how other mutable `web` overlays are handled.

**PhantomBot/docs/guides/content/developerdocs/toc.json**

- Register the **custom modules** developer guide in the TOC — the in-panel docs browser builds from TOC JSON; new guides are invisible without an entry.

**PhantomBot/docs/guides/content/moduleguides/toc.json**

- Register the **adding custom modules** guide in the TOC — installers discover guides through the published guide index; same rationale as the developer TOC.


gmt2001 (changes requested):
Prefer using `let` in all JS var declarations to strictly control scope and prevent global object creation

resources/web/panel/js/utils/customPanelCards.js
```diff
@@ -0,0 +1,383 @@
...
+    /**
+     * Decides whether a stored {@code modules} value means "enabled". The stock convention
+     * is {@code "true"}/{@code "false"} strings but we accept booleans, {@code 1}/{@code 0},
+     * and {@code "enabled"}/{@code "disabled"} just in case a custom module wrote one of
+     * those instead.
+     *
+     * @param {*} raw value from {@code lookupModulesTableValue}
+     * @returns {boolean}
+     */
+    function modulesDbValueIsEnabled(raw) {
```
Sartharon: Shouldn't this:
a. Use the already present helpers.isTrue()?https://github.com/PhantomBot/PhantomBot/blob/9b551e7e7f4af0493bb485e41d6b6fba049738d9/resources/web/panel/js/utils/helpers.js#L1054
b. Be offloaded to helpers.js so it's globally available as an "Enabled" checker similar to helpers.getModuleStatus()?
https://github.com/PhantomBot/PhantomBot/blob/9b551e7e7f4af0493bb485e41d6b6fba049738d9/resources/web/panel/js/utils/helpers.js#L1074

source/tv/phantombot/httpserver/HTTPPanelAndYTHandler.java
```diff
@@ -89,6 +90,32 @@ public void handleRequest(ChannelHandlerContext ctx, FullHttpRequest req) {
...
+                if (cached.matchesIfNoneMatch(ifNoneMatch)) {
+                    FullHttpResponse notModified = HttpServerPageHandler.prepareHttpResponse(HttpResponseStatus.NOT_MODIFIED);
+                    notModified.headers().set(HttpHeaderNames.ETAG, cached.etag());
+                    com.gmt2001.Console.debug.println("304 " + req.method().asciiName() + ": /panel/custom-manifests.json");
+                    HttpServerPageHandler.sendHttpResponse(ctx, req, notModified);
+                    return;
+                }
+
+                FullHttpResponse res = HttpServerPageHandler.prepareHttpResponse(HttpResponseStatus.OK, cached.bytes(), "custom-manifests.json");
+                res.headers().set(HttpHeaderNames.ETAG, cached.etag());
```
Sartharon: Your ETag is not being used by the browsers as you use `HttpServerPageHandler.prepareHttpResponse()` and `HttpServerPageHandler.sendHttpResponse()` which explictly disable caching. 
https://github.com/PhantomBot/PhantomBot/blob/9b551e7e7f4af0493bb485e41d6b6fba049738d9/source/com/gmt2001/httpwsserver/HttpServerPageHandler.java#L348-L349

You can either set your own `Cache-Control` header here to overwrite that behavior or switch to `HttpServerPageHandler.sendFile()` which reverts to basic `filesize` & `lastmodified` as ETag paramters. One could also modify `HttpServerPageHandler.sendFile()` a little (or overload it) so it correctly supports custom computed ETags which would be the cleaner and shorter solution also allowing for Zero-Copy file transfers. 
https://github.com/PhantomBot/PhantomBot/blob/9b551e7e7f4af0493bb485e41d6b6fba049738d9/source/com/gmt2001/httpwsserver/HttpServerPageHandler.java#L422

source/com/mcawful/CustomPanelManifestCollector.java
```diff
@@ -0,0 +1,1118 @@
...
+    /**
+     * Folder values from a manifest must begin with {@code custom/} (so paths land in the
+     * namespaced custom tree) and contain neither {@code ..} nor {@code \\}, which would let a
+     * malicious manifest escape that namespace or trip Windows path quirks. Used for
+     * {@code nav.folder}.
+     *
+     * @param folder manifest-supplied folder string
+     * @return {@code true} if the folder is safe to forward to the panel page loader
+     */
+    private static boolean isSafeFolder(String folder) {
```
Sartharon, line 1093: Shouldn't we guard against any reserved URI characters for URI paths (folder and file names, thus also applies to `isSafePageFile()`). After all you parse user input here and pure path traversal safeguards might not be sufficient I wonder.

See [RFC3986](https://www.rfc-editor.org/rfc/rfc3986) Section 2

Sartharon:
Any plans to integrate permission controls in to this as well with the existing **PanelUser** concept?

https://github.com/PhantomBot/PhantomBot/tree/master/source/tv/phantombot/panel/PanelUser

source/tv/phantombot/httpserver/HTTPPanelAndYTHandler.java
mcawful: Decided to go with the sendFile modification to allow for custom computed ETags

resources/web/panel/js/utils/customPanelCards.js
mcawful: Went with the existing helper function

source/com/mcawful/CustomPanelManifestCollector.java
mcawful: I've made the manifest collector more robust to protect against possible malicious or bad folder/file names

mcawful:
> Any plans to integrate permission controls in to this as well with the existing **PanelUser** concept?
> 
> https://github.com/PhantomBot/PhantomBot/tree/master/source/tv/phantombot/panel/PanelUser

I have some slight integration, but I would like to grow it out. If you want, I could look into integrating that with this PR.

Sartharon:
> > Any plans to integrate permission controls in to this as well with the existing **PanelUser** concept?
> > https://github.com/PhantomBot/PhantomBot/tree/master/source/tv/phantombot/panel/PanelUser
> 
> I have some slight integration, but I would like to grow it out. If you want, I could look into integrating that with this PR.

As I see it there are three options here:

- No integration
     - Everyone has access to this
- Basic integration
     - Users either get _all_ or _no_ access at all
- Full/Advanced integration
     - Users can be assigned different permissions for each custom module

In any case one of the above should be available. "_Full integration_" is not necessarily needed. "`No integration`" is also alright if you choose so.

- But did you test the behavior of this with a non-default user?
- Does it allow the user access and thus fits the "`No integration`" bullet point?

Improvements or stepping up the latter here can always follow in later PRs or via somebody else 😄

gmt2001 (approved):

PR #3722, opened 2026-05-17
Rhino interpret mode support & misc

Sartharon:
- Update Rhino to [1.9.1](https://github.com/mozilla/rhino/releases/tag/Rhino1_9_1_Release)
- Add _botlogin.txt_ property `rhinointerpretmode` to enable Rhino's interpret mode avoiding ByteCode compilation. This allows the user to either choose
     - `true`: the **memory efficiency** of JIT interpreted scripts
     - `false`: the compiled bytecode's **performance**
         - `rhinointerpretmode` defaults to `false`
- Strip the Script class of any unneeded variables and code.
    - Global variables and base scope is build in the RhinoRuntime upon initialisation
    - `Context` is not worth saving as it's volatile and thread-local. `enterContext()` handles proper context attaching

### DIFF
--- a/source/tv/phantombot/script/RhinoRuntime.java
+++ b/source/tv/phantombot/script/RhinoRuntime.java
@@ -95,6 +95,12 @@ public final class RhinoRuntime {
         return BASESCOPE;
     }
 
+
+    public static String callGlobalExposedScriptMethod(String method, String arg) {
+        Object[] obj = new Object[]{arg};
+        return ScriptableObject.callMethod(FACTORY.enterContext(), GLOBAL, method, obj).toString();
+    }
+
     /**
      * Expose a property to all scripts via the {@code $} api. Scripts can later use {@code $.property} and {@code $.property.function()} to access these
      * @param propertyName The name of the property visible to the scripts
--- a/source/tv/phantombot/script/RhinoRuntime.java
+++ b/source/tv/phantombot/script/RhinoRuntime.java
@@ -16,7 +16,9 @@
  */
 package tv.phantombot.script;
 
+import org.mozilla.javascript.NativeObject;
 import org.mozilla.javascript.RhinoException;
+import org.mozilla.javascript.ScriptableObject;
 import org.mozilla.javascript.StackStyle;
 import org.mozilla.javascript.tools.debugger.Main;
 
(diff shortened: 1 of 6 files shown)


source/tv/phantombot/script/RhinoRuntime.java
```diff
@@ -16,25 +16,30 @@
...
 
+import tv.phantombot.CaselessProperties;
+
 /**
  * Rhino runtime environment configuration and management.
  * @author sartharon
  */
 public final class RhinoRuntime {
     private static final RhinoContextFactory FACTORY = new RhinoContextFactory();
+    private static final NativeObject GLOBAL = new NativeObject(); // $ api used in scripts
```
gmt2001, line 33: _CoPilot Conversation:_

I'll help you find a solution for the failing job. Let me start by retrieving the logs.Now let me get the specific file where the error occurs:## Solution for Failing Job 76419351457

**Problem:** The compilation is failing with the error:
```
error: GLOBAL has private access in RhinoRuntime
return ScriptableObject.callMethod(RhinoRuntime.GLOBAL, method, obj).toString();
```

The issue is on line 56 of `Script.java` where the code attempts to access `RhinoRuntime.GLOBAL`, which is a private field.

**Solution:** Change `RhinoRuntime.GLOBAL` to use a public accessor method instead of direct field access.

Update line 56 in [`Script.java`](https://github.com/PhantomBot/PhantomBot/blob/a37be98dbdf4ea350460495629cc61e5ef9ef588/source/tv/phantombot/script/Script.java) from:

```java
return ScriptableObject.callMethod(RhinoRuntime.GLOBAL, method, obj).toString();
```

To:

```java
return ScriptableObject.callMethod(RhinoRuntime.getGlobal(), method, obj).toString();
```

This assumes `RhinoRuntime` has a public `getGlobal()` method that provides access to the global scope. If that method doesn't exist, you'll need to either:
1. Add a public getter method to `RhinoRuntime`
2. Make `RhinoRuntime.GLOBAL` public instead of private
3. Use an alternative API to access the global Rhino scope

Check your `RhinoRuntime` class to confirm the appropriate public accessor method available.

gmt2001 (changes requested):

source/tv/phantombot/script/RhinoRuntime.java
Sartharon: I've forgot about that `callMethod()` static function in the Script class.
As the function should anyhow called on the parents context and with globally (not script local). It is misplaced in the Script class and should reside in the RhinoRuntime class.

gmt2001 (approved):

PR #3727, opened 2026-05-25
Ensure proper ThreadLocal Rhino Context

Sartharon:
We should ensure we enter a fully populated Rhino Context relevant to the script in scope. This is done by entering the context before JavaScript execution in any form (via Bus **Events**, **JSTimers**, or similar). As the context is to be considered thread local we should as well exit them right after the task (method call, timer run, etc.) has completed. This also ensures proper release of resources attached to the context 

To achieve this this PR ensure our **ContextFactory** allows entering a suitable context with the requested options set

### DIFF
--- a/source/com/gmt2001/JSTimers.java
+++ b/source/com/gmt2001/JSTimers.java
@@ -165,8 +165,11 @@ public class JSTimers {
                 if (this.isCancelled.get()) {
                     return;
                 }
+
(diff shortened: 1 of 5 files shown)


gmt2001 (dismissed):

Sartharon:
Branch rebased to allow for merge

PR #3731, opened 2026-05-28
Update GPL3 with fresh copy

Sartharon:
GPL has been modified slightly over the year

Update it with a fresh copy from [GNU](https://www.gnu.org/licenses/gpl-3.0.txt) and fill the placeholders


Sartharon:
~I might be wrong here with a 2015 development start. I will fix it if it is the wrong starting year of this project~

Nvm I was silly it's all over in the code

gmt2001 (approved):

